PR #34, opened 2026-04-18
Issue #23 Fix: Use logged in hf token using `huggingface_hub.get_token()` instead of requiring env variable

### LINKED ISSUE #23: Use logged in hf token using `huggingface_hub.get_token()` instead of requiring env variable
A more robust way to get the logged in HF user would to be just:

```py
from huggingface_hub import get_token

token = get_token()
print(token) 
```

which will read the token, instead of requiring it to be set via an env variable.

vortezwohl:
## Summary
- Add a shared Hugging Face token helper that keeps `HF_TOKEN` as the first local token source and falls back to `huggingface_hub.get_token()`
- Reuse the helper in CLI and backend session creation so users logged in with `huggingface-cli login` do not need to set `HF_TOKEN`
- Update the README quick start notes to document the login-cache fallback while keeping the existing environment variable flow

## Test plan
- [x] Run `python -m py_compile agent\utils\hf_auth.py agent\main.py backend\routes\agent.py`
- [x] Verify `HF_TOKEN` takes precedence over `huggingface_hub.get_token()`
- [x] Confirm README changes are limited to the HF token setup note

Fixes #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,10 @@ Create a `.env` file in the project root (or export these in your shell):
 
 ```bash
 ANTHROPIC_API_KEY=<your-anthropic-api-key> # if using anthropic models
-HF_TOKEN=<your-hugging-face-token>
+HF_TOKEN=<your-hugging-face-token> # optional if you've run huggingface-cli login
 GITHUB_TOKEN=<github-personal-access-token> 
 ```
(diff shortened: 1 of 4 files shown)


vortezwohl:
conflict resolved

akseljoonas:

lewtun:

Current main already resolves CLI/router tokens through huggingface_hub.get_token() in agent/core/hf_tokens.py.

PR #42, opened 2026-04-21
Fix duplicate status display for parallel research agents

lee101:
## Summary
- When multiple research sub-agents run in parallel, they showed identical status in both CLI and web UI because they shared a single global display instance
- Replaced the singleton `SubAgentDisplay` with a `SubAgentDisplayManager` that tracks each agent independently by unique `agent_id` (derived from task hash)
- Each research agent now shows its own labeled status line with distinct tool calls, token counts, and timing

## Changes
- **Backend (CLI)**: `SubAgentDisplayManager` renders all active agents together with per-agent headers, stats, and rolling tool call logs
- **Backend (research_tool)**: Each invocation generates a unique `agent_id` + descriptive `label` passed through `tool_log` events
- **Frontend**: Store tracks `researchAgents` map (keyed by `agent_id`) instead of flat `researchSteps`/`researchStats`, so the UI renders per-agent steps with labels

## Test plan
- [ ] Run 3+ parallel research tasks and verify each shows distinct status in CLI
- [ ] Verify web UI shows separate labeled steps for each concurrent research agent
- [ ] Verify single research agent still works correctly (backward compat)
- [ ] Verify research completion clears display correctly for each agent independently

### DIFF
--- a/agent/main.py
+++ b/agent/main.py
@@ -326,7 +326,9 @@ async def event_listener(
                 tool = event.data.get("tool", "") if event.data else ""
                 log = event.data.get("log", "") if event.data else ""
                 if log:
-                    print_tool_log(tool, log)
+                    agent_id = event.data.get("agent_id", "") if event.data else ""
+                    label = event.data.get("label", "") if event.data else ""
+                    print_tool_log(tool, log, agent_id=agent_id, label=label)
             elif event.event_type == "tool_state_change":
                 pass  # visual noise — approval flow handles this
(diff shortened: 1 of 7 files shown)


akseljoonas:
Thanks for kicking this off! I've opened #49 which builds on your `SubAgentDisplayManager` / per-agent store scaffolding and adds the fixes and UX polish needed to ship it — you're credited as Co-Authored-By on the squashed commit. Closing this in favor of #49.

PR #47, opened 2026-04-22
ollama support

kanpuriyanawab:
As of now ml-intern uses models via litellm which requires paid anthropic or openai key.
This PR adds support for using models via Ollama for harness layer.


kanpuriyanawab:
closing this as a PR already exists.

PR #48, opened 2026-04-22
Rebrand package from hf-agent to ml-intern

akseljoonas:
## Summary
- Rename the distribution name in `pyproject.toml` from `hf-agent` to `ml-intern` (and update the placeholder description and the `[all]` self-reference).
- Regenerate `uv.lock` so the package key matches.
- `uv sync` now shows `Built ml-intern@...` instead of `Built hf-agent@...`.

## Not changed
- `agent/config.py` still references the HF dataset `akseljoonas/hf-agent-sessions`. That's a remote dataset path — renaming here needs the dataset itself renamed on HF, so I left it alone. Let me know if you want that migrated too.

## Test plan
- [ ] `rm -rf .venv && uv sync` — build line reads `ml-intern`, not `hf-agent`.
- [ ] `uv run ml-intern` still launches the CLI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
-name = "hf-agent"
+name = "ml-intern"
 version = "0.1.0"
-description = "Add your description here"
+description = "ML Intern — an AI agent for machine learning engineering on Hugging Face"
 readme = "README.md"
 requires-python = ">=3.11"
 dependencies = [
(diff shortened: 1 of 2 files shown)


akseljoonas:
Closing — distribution rename only affects the one-time build log and isn't worth the migration cost (existing users would need `uv tool install -e . --force`). Script command is already `ml-intern` via `[project.scripts]`.

PR #60, opened 2026-04-22
Draft: Add new providers via adapters

andrejvysny:
## Summary
- builds on top of #55 (provider adapter refactor)
- adds new provider adapters for LM Studio, OpenRouter, OpenCode Zen, and OpenCode Go
- adds session-scoped model switching, local LM Studio discovery, and restore-summary UX fixes

## Verification
- `python3 -m py_compile agent/core/provider_adapters.py agent/core/llm_params.py`
- `uv run --with pytest pytest tests/test_provider_adapters.py`

PR #55, opened 2026-04-22
feat: Refactor existing LLM resolution into provider adapters

andrejvysny:
## Summary
- refactor the existing Anthropic, OpenAI, and Hugging Face router LiteLLM param resolution into small provider adapter classes
- keep existing CLI suggestions and web model/session behavior unchanged while reusing shared model-id validation in the CLI switcher
- centralize provider error classification so the CLI, `/model`, session errors, and `/health/llm` show clean auth/quota/model-not-found messages instead of raw provider traces
- add focused regression tests for adapter parity, validation edge cases, and provider error handling

## Verification
- `uv run --with pytest --with pytest-asyncio pytest -q`
- `uv run ml-intern "Say Hello!" --model "MiniMaxAI/MiniMax-M2.7"`
- `ANTHROPIC_API_KEY= uv run ml-intern "Say Hello!" --model "anthropic/claude-opus-4-6"`
- `OPENAI_API_KEY= uv run ml-intern "Say Hello!" --model "openai/gpt-5"`
- `uv run ml-intern "Say Hello!" --model "fake-org/not-a-real-model-12345"`

PR #68, opened 2026-04-23
Add local OpenAI-compatible backend routing

### LINKED ISSUE #67: Feature request: Allow routing to local OpenAI-compatible backends (Ollama / vLLM / LM Studio)
## Summary
Today `agent/core/llm_params.py::_resolve_llm_params` hard-routes every
model id into one of three branches:

- `anthropic/...` → Anthropic API
- `openai/...` → OpenAI API (no `api_base` override)
- everything else → forced onto `https://router.huggingface.co/v1`
  with an HF token

There's no escape hatch for a local OpenAI-compatible server (Ollama,
vLLM, LM Studio, llama.cpp's server, etc.), even though `litellm`
itself supports all of them out of the box.

## Motivation
Smaller local models almost certainly won't match frontier-model
quality on this agent loop, and that's fine — the value is in opening
up more room for experimentation:

- Cheap / unlimited iteration without per-token costs
- Offline use and stronger privacy for sensitive datasets / papers
- A testbed for the community to benchmark open-weights models on the
  same agent harness
- Lower barrier to entry for students and hobbyists who want to study
  the agent end-to-end

## Suggested shape
Something minimally invasive, e.g.:

- recognize an `ollama/...` prefix and pass it through to litellm, or
- honor an optional `api_base` field in `main_agent_config.json` /
  `LITELLM_API_BASE` env var on the "fallback" branch instead of
  hard-coding the HF router URL.

Related to #56, but specifically about *local* execution rather than
hosted open-source model APIs.

Happy to test or send a PR if this is on the roadmap.

Pro777:
## Summary

Implements the narrow local-backend slice of `#67` without pulling in the larger provider-registry work from `#55` / `#60`.

This PR adds direct routing for these OpenAI-compatible prefixes:

- `ollama/<model>`
- `lm_studio/<model>`
- `vllm/<model>`
- `openai-compat/<model>`

Each prefix resolves to a configurable `api_base` + API key env var while keeping the existing Hugging Face Router fallback path unchanged.

## Why this shape

`#67` asks for an escape hatch to local OpenAI-compatible backends. There is already broader work in `#55` / `#60` around provider registries and additional hosted providers.

This PR intentionally stays smaller:

- no new provider catalog
- no frontend model-picker changes
- no session-manager or restore-summary changes
- no Rowan-specific defaults

It only teaches the current resolver and CLI validation layer about local/OpenAI-compatible prefixes.

## New prefixes

- `ollama/` -> `OLLAMA_BASE_URL` (default `http://localhost:11434/v1`)
- `lm_studio/` -> `LMSTUDIO_BASE_URL` (default `http://127.0.0.1:1234/v1`)
- `vllm/` -> `VLLM_BASE_URL` (default `http://127.0.0.1:8000/v1`)
- `openai-compat/` -> `OPENAI_COMPAT_BASE_URL` (default `http://127.0.0.1:8000/v1`)

The local prefixes set sensible default API keys for backends that typically ignore auth (`ollama`, `lm-studio`, `EMPTY`) while still allowing override via env.

## Tests

Ran:

```bash
uv run --with pytest pytest tests/ -v --tb=short
```

Result:

- `5 passed`

The new tests cover:

- Ollama param resolution
- LM Studio param resolution
- vLLM param resolution
- generic OpenAI-compatible param resolution
- `/model` validation for the new local prefixes

Closes #67.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ Create a `.env` file in the project root (or export these in your shell):
 
 ```bash
 ANTHROPIC_API_KEY=<your-anthropic-api-key> # if using anthropic models
+OLLAMA_BASE_URL=http://localhost:11434/v1  # if using ollama/ models
+LMSTUDIO_BASE_URL=http://127.0.0.1:1234/v1 # if using lm_studio/ models
+VLLM_BASE_URL=http://127.0.0.1:8000/v1     # if using vllm/ models
+OPENAI_COMPAT_BASE_URL=http://127.0.0.1:8000/v1 # generic OpenAI-compatible backend
 HF_TOKEN=<your-hugging-face-token>
 GITHUB_TOKEN=<github-personal-access-token> 
 ```
@@ -50,6 +54,9 @@ ml-intern "fine-tune llama on my dataset"
 
 ```bash
 ml-intern --model anthropic/claude-opus-4-6 "your prompt"
+ml-intern --model ollama/llama3.1:8b "your prompt"
+ml-intern --model lm_studio/google/gemma-4-e4b "your prompt"
+ml-intern --model vllm/meta-llama/Llama-3.1-8B-Instruct "your prompt"
 ml-intern --max-iterations 100 "your prompt"
 ml-intern --no-stream "your prompt"
 ```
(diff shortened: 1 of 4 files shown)


akseljoonas:

PR #83, opened 2026-04-23
fix: validate URL origin in fetch_hf_docs to prevent SSRF

JasonOA888:
## Summary

`fetch_hf_docs` accepted arbitrary URLs from LLM-generated arguments without origin validation. Since the request carries the user's HF Bearer token (`Authorization: Bearer {hf_token}`) and uses `follow_redirects=True`, a crafted prompt could trick the LLM into sending the token to an attacker-controlled server.

## Fix

Add `_is_allowed_doc_url()` which restricts requests to `huggingface.co`, `hf.co`, and `gradio.app` — the three origins this tool is actually designed to fetch from.

The check runs before the HTTP request is made, so no auth headers ever leave allowed origins.

## Before

```python
url = arguments.get("url", "")
if not url:
    return "Error: No URL provided", False
# ... directly fetches url with Bearer token
```

## After

```python
url = arguments.get("url", "")
if not url:
    return "Error: No URL provided", False
if not _is_allowed_doc_url(url):
    return ("Error: URL not allowed. ...", False)
# ... only reaches here for allowed origins
```

## Testing

- Syntax validated (`ast.parse` clean)
- Allowlist covers all existing doc endpoints (huggingface.co/docs/*, gradio.app/llms.txt)
- Subdomains supported via `host.endswith()` check

### DIFF
--- a/agent/tools/docs_tools.py
+++ b/agent/tools/docs_tools.py
@@ -12,6 +12,7 @@ from whoosh.analysis import StemmingAnalyzer
 from whoosh.fields import ID, TEXT, Schema
 from whoosh.filedb.filestore import RamStorage
 from whoosh.qparser import MultifieldParser, OrGroup
+from urllib.parse import urlparse
 
 # ---------------------------------------------------------------------------
 # Configuration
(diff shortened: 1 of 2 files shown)


akseljoonas:

lewtun:

PR #88, opened 2026-04-23
fix: thread thinking_blocks through history so extended thinking survives tool turns

### LINKED ISSUE #87: Extended thinking silently dropped on tool-continuation turns (thinking_blocks not threaded through history)
## Symptom

`ml-intern` requests extended thinking via `thinking={"type": "adaptive"}` (see `agent/core/llm_params.py:153`), but the assistant messages it stores in history never include `thinking_blocks` or `reasoning_content`. On every tool-continuation turn, LiteLLM logs:

```
LiteLLM:WARNING: Dropping 'thinking' param because the last assistant message
with tool_calls has no thinking_blocks. The model won't use extended thinking
for this turn.
```

This silently disables extended thinking for ~every turn after the first tool call — effectively turning Opus into non-thinking-mode Opus for the bulk of an agent's work, which is a meaningful reasoning-quality degradation for what ml-intern is used for.

## Reproduction

Any agent run that goes through a tool call:

```
ml-intern "run bash 'date +%Y'. then tell me the year"
```

…produces the warning on turn 2. With a thinking-worthy prompt, the model *does* produce thinking on turn 1, but it's discarded before turn 2, so the warning still fires.

## Root cause

In `agent/core/agent_loop.py`:

1. `_call_llm_streaming` and `_call_llm_non_streaming` don't capture `thinking_blocks` / `reasoning_content` from responses.
2. `LLMResult` has no fields for them.
3. The three `Message(role="assistant", ...)` construction sites in the loop drop the thinking state entirely.

So the next `acompletion` call sees a tool-call-bearing assistant message with no `thinking_blocks`, and LiteLLM strips the `thinking` param to avoid an API error.

## Proposed fix

~30 lines, no behavior change for non-thinking models:

- Add `thinking_blocks` + `reasoning_content` fields to `LLMResult`.
- **Streaming path**: collect raw chunks during iteration; after the stream finishes, call `litellm.stream_chunk_builder(chunks)` to get a reassembled `ModelResponse`, and pull `message.thinking_blocks` / `message.reasoning_content` off it. Best-effort — wrap in try/except so unfamiliar providers just fall back to no thinking reassembly.
- **Non-streaming path**: read them directly off `response.choices[0].message`.
- Attach both to every `Message(role="assistant", ...)` in the loop (truncation-hint site, no-tool-calls site, with-tool-calls site).

Verified locally: `stream_chunk_builder` handles Anthropic adaptive-thinking deltas correctly (1 thinking block + full `reasoning_content` reassembled from the streamed chunks); the warning disappears on turns where the model actually produced thinking; trivial prompts where adaptive thinking legitimately skips still show the warning because there's genuinely nothing to attach — which is semantically correct.

PR to follow.

triscacezar-droid:
Fixes #87.

## Problem

`ml-intern` requests extended thinking via `thinking={"type": "adaptive"}` (`agent/core/llm_params.py:153`), but the assistant messages it stores in history never carry the `thinking_blocks` / `reasoning_content` that the model produced. On every tool-continuation turn LiteLLM logs:

```
LiteLLM:WARNING: Dropping 'thinking' param because the last assistant message
with tool_calls has no thinking_blocks. The model won't use extended thinking
for this turn.
```

This silently disables extended thinking for ~every turn after the first tool call — effectively turning Opus into non-thinking-mode Opus for the bulk of an agent's work.

## Root cause

In `agent/core/agent_loop.py`:
1. `_call_llm_streaming` and `_call_llm_non_streaming` don't capture `thinking_blocks` / `reasoning_content` from responses.
2. `LLMResult` has no fields for them.
3. All three `Message(role="assistant", ...)` construction sites in the loop drop the thinking state.

## Fix

- **`LLMResult`**: new optional `thinking_blocks` + `reasoning_content` fields.
- **Streaming path**: collect raw chunks during iteration, then reassemble with `litellm.stream_chunk_builder(chunks)` after the stream ends and pull thinking state off the reassembled message. Wrapped in try/except so unfamiliar providers fall back silently.
- **Non-streaming path**: read them directly off `response.choices[0].message`.
- **Main loop**: attach both fields to every `Message(role="assistant", ...)` construction (truncation site, no-tool-calls terminal site, with-tool-calls site).

## Verification

Tested locally with `anthropic/claude-opus-4-6` + adaptive thinking:

- `stream_chunk_builder` correctly reassembles 1 thinking block + full `reasoning_content` from streamed chunks.
- LiteLLM warning **disappears** on turns where the model actually produced thinking.
- Trivial prompts where adaptive thinking legitimately skips still show the warning because there's genuinely nothing to attach — semantically correct.

## Compatibility

No behavior change for non-thinking models — all new fields default to `None` and are only populated when present on the response. Safe for HF router / OpenAI / any provider `stream_chunk_builder` doesn't know how to handle (falls back to `None`).

### DIFF
--- a/agent/core/agent_loop.py
+++ b/agent/core/agent_loop.py
@@ -9,7 +9,7 @@ import os
 import time
 from dataclasses import dataclass, field
 
-from litellm import ChatCompletionMessageToolCall, Message, acompletion
+from litellm import ChatCompletionMessageToolCall, Message, acompletion, stream_chunk_builder
 from litellm.exceptions import ContextWindowExceededError
 
 from agent.config import Config
@@ -396,6 +396,11 @@ class LLMResult:
     token_count: int
     finish_reason: str | None
     usage: dict = field(default_factory=dict)
+    # Extended-thinking state. Must be threaded back into the next assistant
+    # Message for Anthropic models, otherwise LiteLLM strips the `thinking`
+    # param on tool-continuation turns and we lose extended thinking.
+    thinking_blocks: list | None = None
+    reasoning_content: str | None = None
 
 
 async def _call_llm_streaming(session: Session, messages, tools, llm_params) -> LLMResult:
@@ -448,11 +453,16 @@ async def _call_llm_streaming(session: Session, messages, tools, llm_params) -> 
     token_count = 0
     finish_reason = None
     final_usage_chunk = None
+    # Keep the raw chunks around so we can reassemble thinking_blocks /
+    # reasoning_content at the end via litellm.stream_chunk_builder. We do
+    # our own content/tool_call accumulation for backwards compatibility.
+    chunks: list = []
 
     async for chunk in response:
         if session.is_cancelled:
             tool_calls_acc.clear()
             break
+        chunks.append(chunk)
 
         choice = chunk.choices[0] if chunk.choices else None
         if not choice:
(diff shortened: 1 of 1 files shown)


akseljoonas:

triscacezar-droid:
Rebased onto current main — `MERGEABLE` again. Conflicts were trivial: main added `usage` tracking to `LLMResult` / the streaming setup, this PR added `thinking_blocks` / `reasoning_content` / `chunks` in the same regions. Both sides preserved, no semantic changes from the version you reviewed. The Claude bot review last week came back "safe to merge, low risk".

@akseljoonas would appreciate another look when you have a moment 🙏 — without this fix, every ml-intern run on Anthropic models silently loses extended thinking after the first tool call.

lewtun:

Thinking-history handling was already covered by merged PRs #143/#150 and then revised by later router-compatibility work.

PR #113, opened 2026-04-25
feat: add Claude Code project-mode support

GuyNachshon:
# feat: add Claude Code project-mode support

## Summary

Adds Claude Code as a second frontend alongside the standalone `ml-intern` CLI. Both share the same tools under `agent/tools/` — no duplication, no changes to the agent runtime.

After this PR, running `claude` from the repo root works:

```bash
claude                                          # interactive
claude -p "fine-tune llama on my dataset"       # headless
```

## What's new

| File | Purpose |
|---|---|
| `packages/mcp_server/server.py` | MCP server that re-exposes `agent/tools/*` to Claude Code via stdio. Uses `mcp.server.lowlevel.Server` to preserve existing JSON schemas verbatim — FastMCP 3.x re-derives schemas from Python type hints, which would lose `oneOf`/operation-discriminated structures. |
| `CLAUDE.md` | System prompt ported from `agent/prompts/system_prompt_v3.yaml`. `plan_tool` references → TodoWrite. `research` tool → Task subagent. |
| `.claude/agents/research.md` | Research subagent (Claude Code's equivalent of the `research_tool` — its own context, restricted to read-only HF tools). |
| `.claude/commands/{ml-intern,research,inspect-dataset,finetune,run-job}.md` | Five slash commands, opinionated to match `CLAUDE.md` methodology. |
| `.claude/hooks/pre_tool_use_approval.py` | Port of `_needs_approval` from `agent/core/agent_loop.py`. Fail-safe: malformed payloads, non-dict `tool_input`, and empty `tool_name` all force a prompt instead of silent allow. |
| `.claude/hooks/session_start_context.py` | Injects HF username (via `huggingface_hub.HfApi.whoami()`) and a local-mode banner so the model knows the user's HF namespace and whether to expect a sandbox. Mirrors `agent/context_manager/manager.py`. |
| `.claude/hooks/session_end_upload.py` | Uploads transcripts to `smolagents/ml-intern-sessions` after running through `agent/core/redact.py::scrub`. Refuses paths outside `~/.claude/` or `$CLAUDE_PROJECT_DIR`. |
| `CLAUDE_CODE_GUIDE.md` | User docs (slash commands, approvals, env knobs, troubleshooting). |
| `.gitignore` | Switch from blanket `.claude/` to specific exclusions (`.claude/settings.local.json`, `__pycache__`) so shared project config is tracked but per-user overrides are not. |

## What's *not* changed

- `agent/` is untouched. The standalone `ml-intern` CLI works exactly as before.
- No new Python deps. The MCP server uses `mcp` and `fastmcp` already in `pyproject.toml`.
- No CI changes (the vendoring-drift guard is in the follow-up plugin PR).

## Behavior parity vs. standalone CLI

| `Config` field | Hook env var | Default |
|---|---|---|
| `yolo_mode` | `ML_INTERN_YOLO` | `0` |
| `confirm_cpu_jobs` | `ML_INTERN_CONFIRM_CPU_JOBS` | `1` |
| `save_sessions` | `ML_INTERN_SAVE_SESSIONS` | `1` |
| `session_dataset_repo` | `ML_INTERN_SESSION_REPO` | `smolagents/ml-intern-sessions` |
| `--local` | `ML_INTERN_LOCAL_MODE` | `0` |

## What's *intentionally* not ported

- Slash commands `/help`, `/compact`, `/model`, `/yolo`, `/effort`, `/status`, `/undo` — Claude Code has natives or these are model-loop concepts (`/effort`, `/model` probe) that don't apply.
- `record_llm_call`, `record_hf_job_*`, `record_sandbox_*`, `HeartbeatSaver` — Claude Code's transcript persistence covers the externally-visible behavior (don't lose data on long turns).
- `auto_save_interval`, `heartbeat_interval_s`, `max_iterations`, `reasoning_effort`, `prompt_caching`, `effort_probe`, `model_switcher`, `hf_router_catalog` — Claude Code provides equivalents.
- `private_hf_repo_tools` — already disabled upstream (`agent/core/tools.py:55-58`).

## Test plan

- [x] `uv run python -m packages.mcp_server.server < /dev/null` boots cleanly (15 tools register: 10 HF tools + 4 sandbox tools + `sandbox_create`).
- [x] Approval hook fail-safes on malformed stdin, empty `tool_name`, non-dict `tool_input` (all force `ask`).
- [x] Approval hook auto-approves CPU jobs with `push_to_hub` when `ML_INTERN_CONFIRM_CPU_JOBS=0`.
- [x] Approval hook always prompts for GPU jobs.
- [x] Approval hook surfaces `from_pretrained` without `push_to_hub` warning in the prompt reason.
- [x] SessionStart hook resolves HF username when `HF_TOKEN` is set; reports specific failure mode (no token vs whoami HTTP error vs other) when not.
- [x] SessionEnd hook redacts `hf_…`, `sk-ant-…`, `sk-…`, `gh*_…`, `Bearer …`, and `KEY=value` exports before upload (verified against planted secrets).
- [x] SessionEnd hook refuses paths outside `~/.claude/` or `$CLAUDE_PROJECT_DIR`.
- [x] Standalone `ml-intern` CLI continues to work unchanged.
- [x] Existing `tests/unit/` (excluding pre-broken `test_user_quotas.py` which lacks `pytest-asyncio`) passes 46/46.

## Anticipated review feedback

- **"Why low-level `mcp.server.lowlevel.Server` instead of FastMCP?"** Documented in the docstring at top of `server.py`. FastMCP 3.x derives input schemas from Python type hints; the existing `*_TOOL_SPEC["parameters"]` JSON schemas in `agent/tools/` use `oneOf`/operation-discriminated structures that don't round-trip through that.
- **"Why three hooks instead of one?"** They fire on different lifecycle events (SessionStart / PreToolUse / SessionEnd). Merging would couple unrelated concerns.
- **"What about telemetry?"** Claude Code writes the transcript continuously, so the externally-visible "don't lose trace data on long turns" behavior is covered. The internal `Event` channel was a `Session`-scoped abstraction without a Claude Code analog.
- **"`.gitignore` change might affect existing users."** Only affects users who had local-only files in `.claude/` (rare — that directory was previously fully ignored, so anything there was either accidental or intentional opt-out). The new rules track everything except `settings.local.json` (Claude Code's standard local-only override file) and runtime caches.

## Follow-up

The plugin packaging (separate `plugin/` directory, vendored library, marketplace manifest) is the next PR. This one stops at "the repo works as a Claude Code project."

### DIFF
--- a/.claude/agents/research.md
+++ b/.claude/agents/research.md
@@ -1,0 +1,119 @@
+---
+name: research
+description: Use proactively before writing any ML implementation code. Mines the literature to find the best training recipes backed by published results, then validates them with working code and current docs. The main agent uses these findings to implement the actual solution. Spawn with a specific brief — name anchor papers or arxiv IDs when you have them.
+tools: Read, Bash, Grep, Glob, WebFetch, mcp__ml-intern-tools__explore_hf_docs, mcp__ml-intern-tools__fetch_hf_docs, mcp__ml-intern-tools__hf_papers, mcp__ml-intern-tools__hf_inspect_dataset, mcp__ml-intern-tools__github_find_examples, mcp__ml-intern-tools__github_list_repos, mcp__ml-intern-tools__github_read_file, mcp__ml-intern-tools__hf_repo_files
+---
+
+You are a research sub-agent for an ML engineering assistant. Your primary job: mine the literature to find the best training recipes — then back them up with working code and up-to-date documentation. The main agent will use your findings to implement the actual solution.
+
+# Start from the literature
+
+Your default approach is a deep literature crawl. Do not start from docs or example scripts — start from papers. Papers contain the results, and results tell you what actually works.
+
+## The crawl
+
+1. **Find anchor papers**: Search for the task/domain. Identify the landmark paper(s) — high citations, recent, or both.
+2. **Crawl the citation graph**: Use `citation_graph` on the anchor paper(s). Look DOWNSTREAM (papers that cite it) — these are the ones that built on it, improved it, or applied it to new domains. Prioritize recent papers and papers with many citations.
+3. **Read methodology sections**: For the most promising papers (strong results, recent, relevant), use `read_paper` with section parameter to read sections 3, 4, 5 (Methodology, Experiments, Results — not the abstract). Extract:
+   - The exact dataset(s) used (name, source, size, any filtering/preprocessing)
+   - The training method and configuration (optimizer, lr, schedule, epochs, batch size)
+   - The results those choices produced (benchmark scores, metrics, comparisons)
+4. **Attribute results to recipes**: This is the critical step. Every finding must link a RESULT to the RECIPE that produced it. "Dataset X + method Y + lr Z → score W on benchmark V" is useful. "They used SFT" is not.
+5. **Validate datasets**: For the most promising datasets, check if they exist on HF Hub with `hf_inspect_dataset`. Verify format matches the training method. Report if it doesn't.
+6. **Find code**: Now find working implementation code via `github_find_examples` and `github_read_file`. Use docs (`explore_hf_docs`, `fetch_hf_docs`) to fill in API details.
+
+## When to go deeper
+
+- If the anchor paper is old (>1 year), its citation graph is your main source — the downstream papers will have better methods.
+- If a downstream paper reports significantly better results, crawl ITS citation graph too.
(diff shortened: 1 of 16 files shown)


fglogan:
closed per maintainer request

lewtun:

The Claude Code project-mode branch is stale, blocked, and already marked for closure by maintainer request.

PR #116, opened 2026-04-25
Add Slack gateway

lewtun:
## Summary

- Adds a Slack-backed notification gateway for server-configured destinations.
- Adds the `notify` tool so agents can send messages to explicitly allowed destinations.
- Wires CLI and headless runs to send approval, error, and turn-completion notifications, with completion notifications deferred until the final response is rendered.
- Adds Slack user/default config from environment variables while preserving explicit `messaging.auto_event_types` configuration.
- Expands turn-completion summaries to a Slack-sized 39k character cap.
- Fixes native Anthropic extended-thinking replay for streaming responses by rebuilding the final signed thinking block from LiteLLM chunks instead of replaying raw thinking deltas.

## Why the Anthropic fix is included

Native Anthropic models can return signed `thinking` blocks when extended thinking is enabled. In streaming mode, LiteLLM surfaces those as separate text and signature deltas. Replaying the raw deltas as complete history blocks can make Anthropic reject the next request with `Invalid signature in thinking block`.

The PR now stores the rebuilt signed block and retries once without stale thinking metadata if an existing session already contains invalid replay data.

See the instructions in `README.md` for a Slack configuration example.


agent/core/session.py
```diff
@@ -12,10 +12,14 @@
 
 from agent.config import Config
 from agent.context_manager.manager import ContextManager
+from agent.messaging.gateway import NotificationGateway
+from agent.messaging.models import NotificationRequest
 
 logger = logging.getLogger(__name__)
 
 _DEFAULT_MAX_TOKENS = 200_000
+_LEGACY_AUTO_EVENT_TYPES = {"approval_required", "error"}
```
lewtun: Why are these things called legacy? We don't care about backwards compatibility :P

agent/core/session.py
```diff
@@ -79,31 +83,39 @@ def __init__(
...
+        session_id: str | None = None,
         user_id: str | None = None,
     ):
         self.hf_token: Optional[str] = hf_token
         self.user_id: Optional[str] = user_id
         self.tool_router = tool_router
         self.stream = stream
         tool_specs = tool_router.get_tool_specs_for_llm() if tool_router else []
+        effective_config = config or Config(
+            model_name="bedrock/us.anthropic.claude-sonnet-4-5-20250929-v1:0",
```
lewtun: Why is this hard-coded?

agent/core/session.py
lewtun: What is this legacy stuff? We don't need to be backwards compat

agent/core/session.py
lewtun: Why is this hard-coded?

agent/core/session.py
```diff
@@ -79,31 +82,39 @@ def __init__(
...
```
lewtun: Why is this hardcoded?

agent/core/agent_loop.py
lewtun, line 1: @akseljoonas these changes are unrelated to the current PR but are needed to deal with an issue on `main` when  working with Anthropic models

PR #120, opened 2026-04-25
fix: enable dotenv override for environment variables

Sumit189:
## Problem

When a user has a stale or invalid API key exported in their shell profile
(e.g. `export OPENAI_API_KEY=...` in `~/.zshrc`), the project's `.env` file
is silently ignored because `load_dotenv()` defaults to `override=False`.

This makes the error message misleading, it suggests adding the key to
`.env`, but that doesn't actually fix anything when a shell export exists:
`Error: Authentication failed — your API key is missing or invalid. ... You can also add it to a .env file in the project root.`

The user ends up stuck: the key works with `curl`, the `.env` looks correct,
but `ml-intern` keeps rejecting it with no indication that a shell variable
is shadowing the `.env` value.

## Fix
Pass `override=True` to the project-root `load_dotenv()` call so the `.env`
file is the source of truth for API keys, matching the behavior the error
message already promises.
```diff
-    load_dotenv(_PROJECT_ROOT / ".env")
+    load_dotenv(_PROJECT_ROOT / ".env", override=True)
     load_dotenv(override=False)
```


fglogan:
closed per maintainer request

lewtun:

The branch is conflicted and its environment-variable docs/behavior are stale after the HF_TOKEN/HF Router shift.

PR #121, opened 2026-04-25
Fix CLI rendering corruption and split CLI/frontend model defaults

akseljoonas:
## Summary
- stop the CLI research overlay from running a background redraw ticker that races with streamed markdown output
- keep research sub-agent routing on direct Anthropic when the active model is direct Anthropic, and mute LiteLLM warning spam in the interactive UI
- split model defaults into explicit CLI and frontend config files, and align the frontend Claude constant with the backend's Bedrock default
- restore missing recovery helpers and patch dangling tool-result handling, doom-loop polling detection, sandbox reuse messaging, and async test support

## Testing
- `python -m compileall agent backend`
- `./frontend/node_modules/.bin/tsc -p frontend/tsconfig.json --noEmit`
- `UV_CACHE_DIR=/tmp/uv-cache uv run --with pytest --with pytest-asyncio python -m pytest -q`


akseljoonas:
Validated the remaining review points before changing code:

- Fixed: removed the orphaned `configs/main_agent_config.json`.
- Fixed: shortened `_LLM_RATE_LIMIT_RETRY_DELAYS` to the two entries the retry loop actually consumes, while keeping the effective retry budget above 60s.
- Not applied: the Bedrock Sonnet 4.6 `-v1` suggestion is incorrect. The current AWS Bedrock model card lists `anthropic.claude-sonnet-4-6` with geo/global IDs like `us.anthropic.claude-sonnet-4-6`, so leaving `bedrock/us.anthropic.claude-sonnet-4-6` unchanged is intentional.

Local verification after the update: `87 passed` with `UV_CACHE_DIR=/tmp/uv-cache uv run --with pytest --with pytest-asyncio python -m pytest -q`.

PR #132, opened 2026-04-26
Run jobs under user or org accounts with upgrade / org-pick UX

akseljoonas:
## Summary
- run `hf_jobs` under the caller's own Pro account or a paid org they belong to instead of any stale shared-org assumption
- if a free user has paid orgs, require them to choose the org from a dropdown before continuing the approved job tool call
- if the user has no eligible jobs namespace, show an upgrade-or-decline dialog and let the agent continue without `hf_jobs` when declined
- preserve the existing org-join onboarding gate for frontend funnel observability
- record jobs gating and Pro CTA clicks in session telemetry and KPI rollups, while preserving correct `hf_jobs` attribution

## Testing
- `python -m compileall agent backend scripts`
- `./frontend/node_modules/.bin/tsc -p frontend/tsconfig.json --noEmit`
- `UV_CACHE_DIR=/tmp/uv-cache uv run --with pytest --with pytest-asyncio python -m pytest -q`

### DIFF
--- a/agent/core/agent_loop.py
+++ b/agent/core/agent_loop.py
@@ -1137,6 +1137,9 @@ class Handlers:
                     tool_args["script"] = edited_script
                     was_edited = True
                     logger.info(f"Using user-edited script for {tool_name} ({tc.id})")
+                selected_namespace = approval_decision.get("namespace")
+                if selected_namespace and tool_name == "hf_jobs":
+                    tool_args["namespace"] = selected_namespace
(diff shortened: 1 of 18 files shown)


AbijahKaj:
Hey @akseljoonas !
Thanks for this tool and this PR too!
Any idea on when this will be deployed on PROD (hosted version)? I'm unable to use my PRO subscription to launch jobs. Still having the popup not willing to go away even though I have a susbcritpion.

Thanks!

PR #136, opened 2026-04-26
feat(session): include user_id + total_cost_usd in trajectory dump

jagwar:
## Summary

Add `user_id` (the HF username from OAuth, already stored on `AgentSession`) to the `Session` object and propagate it into the trajectory JSON that `save_trajectory_local()` writes and the session uploader sends to the dataset. Also surface a `total_cost_usd` aggregate summed from existing `llm_call` events.

## Why

Today the session dataset exposes \`session_id\` but no user binding, so \$-spend per HF user cannot be derived from the dataset alone. The only fallback is correlating timestamps between Hub access logs and the dataset / Bedrock invocation logs — that approach is noisy and breaks completely for users with long-lived browser tabs (their JWT polls inflate their attribution score independently of actual prompts).

With \`user_id\` in the dump, per-user spend analysis becomes a one-line aggregate over the dataset.

## Changes

- \`agent/core/session.py\`
  - \`Session.__init__\` accepts \`user_id: str | None = None\` and stores it
  - \`get_trajectory()\` includes \`user_id\` and a \`total_cost_usd\` aggregate (summed from the \`llm_call\` events already emitted by \`agent.core.telemetry.record_llm_call\`)
- \`backend/session_manager.py\`: pass \`user_id=user_id\` when constructing \`Session\` in \`SessionManager.create_session\`. \`AgentSession.user_id\` is already populated from the OAuth context, no other plumbing needed.

Total diff: 2 files, +13 / -1.

## Test plan

- [ ] Session uploaded to the dataset shows non-null \`user_id\` for OAuth-authenticated sessions, \`null\` (or \`"dev"\`) for CLI / dev mode
- [ ] \`total_cost_usd\` matches the sum of \`cost_usd\` across the session's \`llm_call\` events
- [ ] Existing fields (\`session_id\`, \`session_start_time\`, \`messages\`, \`events\`, \`tools\`) unchanged

## Privacy note

\`user_id\` is the public HF username (already in HF audit logs and in the access logs that ship to ES). Cost is per-session aggregate. No PII beyond what's already discoverable from the user's Space activity.

### DIFF
--- a/agent/core/session.py
+++ b/agent/core/session.py
@@ -79,8 +79,10 @@ class Session:
         hf_token: str | None = None,
         local_mode: bool = False,
         stream: bool = True,
+        user_id: str | None = None,
     ):
         self.hf_token: Optional[str] = hf_token
+        self.user_id: Optional[str] = user_id
         self.tool_router = tool_router
         self.stream = stream
         tool_specs = tool_router.get_tool_specs_for_llm() if tool_router else []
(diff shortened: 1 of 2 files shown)


lewtun (approved):
Thanks for the improvements! LGTM

Issue #61, opened 2026-04-22
Cost, stability, and loop-control improvements


AbijahKaj:
Observed in a live session: parent agent (`anthropic/claude-opus-4-6`, `reasoning_effort=high`) spawned parallel `research` subagents that each climbed past 100k context tokens with no cost guard, no caching, and no concurrency cap. Anthropic 529 overloads triggered silent re-spawns, producing what looked like a runaway loop in the TUI.

### P0 — money-bleed fixes (land first)

- [ ] **Enable Anthropic prompt caching.** No `cache_control` markers anywhere in the repo; cache hit rate is 0%. Add `{"type": "ephemeral"}` breakpoints on (1) system prompt, (2) tool specs block, (3) research subagent system prompt, (4) rolling stable conversation prefix after compaction. Expected ~5–10× input-token cost reduction on tool-heavy loops.
- [ ] **Lower `max_iterations`.** Defaults are too high for Opus 4.6:
  - `agent/config.py` — `max_iterations: int = 300` → 40 (or make it model-aware).
  - `agent/tools/research_tool.py` — `max_iterations = 60` → 20.
- [ ] **Cap concurrent research subagents per turn.** In `agent/core/agent_loop.py` around the `asyncio.gather(...)` at line ~695, refuse (or serialize) more than 1 `research` call at a time. Currently the LLM can fire N parallel research subagents and each runs a full 60-iteration loop.
- [ ] **Default the research model to something cheaper than Sonnet 4.6.** `_get_research_model` in `research_tool.py` picks Sonnet for all Anthropic parents. Haiku 4.x (or an HF router model) is fine for literature crawls; keep Opus only for the orchestrator.

### P1 — stop the "looping out of control" pattern

- [ ] **Circuit-break on repeated subagent failures.** `research_handler` returns `"Research agent LLM error: ..."` on 529/timeouts; the parent happily calls `research` again. Track consecutive failures per session and short-circuit after 2 with a terminal error the LLM cannot retry around.
- [ ] **Exponential backoff on 529/overload in `_call_llm_streaming`.** Current `_LLM_RETRY_DELAYS = [5, 15, 30]` is fine for connection blips but not for sustained overload. Add jitter and raise max delay; consider failing fast when upstream is in a multi-minute overload.
- [ ] **Strengthen `doom_loop.py`.** It hashes `name + md5(args)` so retried `research` calls with slightly different `task` strings never match. Add:
[...]

PR #141, opened 2026-04-27
Lower default max_iterations to prevent runaway cost

lblommesteyn:
## Summary

Addresses the P0 cost-bleed items in #61.

The current defaults (300 for the main agent, 60 for the research subagent) allow a single turn to make hundreds of Opus calls with extended thinking, easily burning $50+ before any user-visible signal. The context budget (170k/190k tokens) is the real limiter for well-behaved runs; `max_iterations` is a safety cap for loops that escape doom-loop detection.

## Changes

- `agent/config.py`: `max_iterations` default 300 → 40
- `agent/tools/research_tool.py`: research subagent `max_iterations` 60 → 20

## Notes

- Both values remain overridable via config for power users who need longer runs
- The comment "context budget is the real limiter" in `research_tool.py` still holds — this just tightens the fallback ceiling

## Test plan

- [ ] Normal task completes well within 40 iterations
- [ ] A deliberately long research task hits the 20-iteration cap and returns whatever it has rather than running indefinitely
- [ ] Users who set `max_iterations: -1` in their config are unaffected

### DIFF
--- a/agent/config.py
+++ b/agent/config.py
@@ -32,7 +32,7 @@ class Config(BaseModel):
     # 0 = disabled. Consumed by agent.core.telemetry.HeartbeatSaver.
     heartbeat_interval_s: int = 60
     yolo_mode: bool = False  # Auto-approve all tool calls without confirmation
-    max_iterations: int = 300  # Max LLM calls per agent turn (-1 = unlimited)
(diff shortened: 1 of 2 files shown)


akseljoonas:
The defaults are there to ensure agent's performance for long running tasks.

PR #154, opened 2026-04-27
fix(bedrock): force tool cachePoint via cache_control_injection_points

jagwar:
## Bug

LiteLLM 1.83.0's Bedrock Converse adapter reads `cache_control` blocks on system content blocks but ignores `cache_control` on individual tool dicts. Only the native Anthropic adapter inspects tool dicts.

Result: the ~16k tokens of tool definitions are re-billed at full input price on every Bedrock turn instead of hitting the cache_read tier.

Confirmed by reading `litellm/llms/bedrock/chat/converse_transformation.py` — the tool path appends a `cachePoint` only when the kwarg `cache_control_injection_points=[{location: tool_config}]` is passed top-level. The `cache_control` on the last tool dict (set by `agent/core/prompt_caching.py`) is silently dropped.

## Fix

Pass `cache_control_injection_points` to the Converse adapter from `_resolve_llm_params` for any `bedrock/...` model.

## Expected impact

Cache-read ratio on Opus 4.6 should jump from ~17% (Bedrock auto-prefix caching only) towards 50-70% on multi-turn sessions, cutting the input bill ~25-35%.

## Test plan

- [ ] Compare CloudWatch `CacheReadInputTokenCount` / `InputTokenCount` ratio on `us.anthropic.claude-opus-4-6-v1` before vs 24h after deploy
- [ ] Verify no regression on `anthropic/claude-...` (untouched branch)

Mirror on HF Space: huggingface.co/spaces/smolagents/ml-intern/discussions/24

Diff: 1 file, +12 / -1.

### DIFF
--- a/agent/core/llm_params.py
+++ b/agent/core/llm_params.py
@@ -166,18 +166,20 @@ def _resolve_llm_params(
         # (``AWS_ACCESS_KEY_ID`` / ``AWS_SECRET_ACCESS_KEY`` / ``AWS_REGION``).
         # The Anthropic thinking/effort shape is not forwarded through Converse
         # the same way, so we leave it off for now.
-        #
+        params: dict = {"model": model_name}
         # ``cache_control_injection_points`` instructs the Converse adapter to
         # append a cachePoint at the end of the tool list. Per-tool
         # ``cache_control`` blocks (set by prompt_caching.py for the Anthropic
         # native path) are otherwise silently dropped by Converse, leaving the
-        # ~16k tokens of tool defs uncached on every Bedrock turn. System-prompt
-        # caching still works via cache_control on system content blocks
-        # (Converse reads those).
-        return {
-            "model": model_name,
-            "cache_control_injection_points": [{"location": "tool_config"}],
-        }
+        # ~16k tokens of tool defs uncached on every Bedrock turn.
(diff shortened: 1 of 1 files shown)


jagwar:
Closing — local test on `bedrock/us.anthropic.claude-opus-4-6-v1` shows the change has no measurable cache benefit on top of the existing system-block `cache_control` set by `prompt_caching.py`.

**Test results (litellm 1.83.0, ~11.9K-token prefix):**

| Scenario | cache_creation | cache_read |
|---|---|---|
| no cache hints | 0 | 0 |
| system `cache_control` only | 11550 | 11550 |
| system + tool inject (this PR) | 11546 | 11546 |
| tool inject only (no system cache) | 0 | 5521 |

**Why:** Bedrock's system `cachePoint` extends through the deterministic prefix (system + tools). Adding a second `cachePoint` at end of `toolConfig.tools` creates an overlapping cache entry but does not extend the cached length. The earlier comment in `llm_params.py` ("~16k tokens of tool defs uncached") was incorrect — tools ARE being cached today via the system breakpoint.

The tool inject would only help in the edge case where system cache_control is absent, which `prompt_caching.py` rules out for any model containing "anthropic" (matches `bedrock/us.anthropic.*`).

A small follow-up PR will fix the misleading comment.

PR #159, opened 2026-04-27
Add web search tool

akseljoonas:
## Summary
- add a Python web_search tool that mirrors Claw Code WebSearch behavior over DuckDuckGo HTML
- register it with the tool router and research sub-agent
- cover DuckDuckGo result parsing, redirect decoding, domain filtering, generic fallback, handler JSON, and registration

## Verification
- pytest tests/unit/test_web_search_tool.py tests/unit/test_malformed_args_recovery.py -q
- python -m compileall agent/tools/web_search_tool.py agent/core/tools.py agent/tools/research_tool.py agent/tools/__init__.py tests/unit/test_web_search_tool.py

### DIFF
--- a/agent/core/tools.py
+++ b/agent/core/tools.py
@@ -50,6 +50,7 @@ from agent.tools.papers_tool import HF_PAPERS_TOOL_SPEC, hf_papers_handler
 from agent.tools.plan_tool import PLAN_TOOL_SPEC, plan_tool_handler
 from agent.tools.research_tool import RESEARCH_TOOL_SPEC, research_handler
 from agent.tools.sandbox_tool import get_sandbox_tools
+from agent.tools.web_search_tool import WEB_SEARCH_TOOL_SPEC, web_search_handler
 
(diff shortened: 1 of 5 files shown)


akseljoonas:
Addressed the review feedback in cf30f5f:

- Wrapped the synchronous search execution in asyncio.to_thread from web_search_handler so the 20s requests timeout does not block the event loop.
- Threaded tool_call_id through web_search_handler into the output tool_use_id, and passed research sub-agent tool call ids through ToolRouter.call_tool.
- Hardened query validation so JSON null/non-string query values fail instead of becoming 'None'.
- Added test coverage for to_thread dispatch, propagated tool_use_id, and non-string query rejection.

Verification:
- pytest tests/unit/test_web_search_tool.py tests/unit/test_malformed_args_recovery.py -q
- python -m compileall agent/tools/web_search_tool.py agent/core/tools.py agent/tools/research_tool.py agent/tools/__init__.py tests/unit/test_web_search_tool.py

PR #166, opened 2026-04-28
Improved local model support

bhargav1000:
## Summary

Adds improved feature-gated local model support for OpenAI-compatible endpoints:

- `ollama/<model>`
- `vllm/<model>`
- `llamacpp/<model>`
- `local://<model>`

This wires local routing through LiteLLM, CLI `/model`, backend validation/catalogs, and the web model selector. Local model UI/API support is gated by `ENABLE_LOCAL_MODELS=true`; local inference servers must already be running.

## Review Follow-up

Addressed review feedback:

- Warn when `reasoning_effort` is passed to local models in non-strict mode.
- Keep strict-mode local `reasoning_effort` rejection explicit.
- Removed redundant whitespace validation.
- Removed fragile `sys.path` injection in local model validation tests.
- Added explicit feature-flag coverage for custom local IDs.
- Documented that local endpoint env vars must remain server-controlled.

## Validation

- `uv --cache-dir /tmp/uv-cache run pytest` - 178 passed, 3 skipped
- `npm run build` from `frontend/` - passed, with existing Vite chunk-size warning
- `python -m py_compile agent/core/llm_params.py backend/model_catalog.py` - passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,8 @@ LLAMACPP_BASE_URL=http://localhost:8001
 LOCAL_LLM_BASE_URL=http://localhost:8000
 ```
 
+Keep these endpoint variables server-controlled. Do not expose them as user-editable web/API inputs; they determine where the backend sends LLM traffic.
+
 For example, with Ollama:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -106,6 +106,82 @@ JSON file:
 }
 ```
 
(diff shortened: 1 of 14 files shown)


fglogan:
closed per maintainer request

lewtun:

Current main already has newer local-model support across docs, routing, and tests.

PR #182, opened 2026-04-29
Azure foundry support added

nikhilvishwanath:
## Summary
Adds support for models deployed on Azure AI Foundry via LiteLLM's `azure/` prefix.

## Changes
- `agent/core/llm_params.py`: Handle `azure/<deployment>` model names, reading credentials from `AZURE_API_KEY`, `AZURE_API_BASE`, and `AZURE_API_VERSION` env vars
- `README.md`: Document Azure AI Foundry setup with required env vars and usage examples

## Usage
Set the following in your `.env`:
AZURE_API_KEY=<<API KEY>>
AZURE_API_BASE=https://<resource>.openai.azure.com/
AZURE_API_VERSION=<<Version>>
## Run 
ml-intern --model azure/<your-deployment-name> "your prompt"

PR #180, opened 2026-04-29
Add TokenRouter model support

aryan5v:
Adds TokenRouter as an OpenAI compatible model target.

What changed
- support tokenrouter/<model> in LiteLLM params
- read TOKENROUTER_API_KEY and optional TOKENROUTER_API_BASE
- add tokenrouter/auto:balance to the CLI and web model lists
- document usage in the README

Verified
- uv run --extra dev pytest tests/unit/test_llm_params.py
- npm run build

[TokenRouter](https://www.tokenrouter.com)
[Docs](https://www.tokenrouter.com/docs)

PR #185, opened 2026-04-29
fix(research): inherit main model's Bedrock region prefix in sub-agent

### LINKED ISSUE #184: Research sub-agent ignores the main model's Bedrock inference-profile prefix
## Bug

When the main model is a `bedrock/` Anthropic profile, the research sub-agent forces a `us.` inference-profile prefix regardless of what the user chose for the main model. This breaks the sub-agent for any user whose AWS account/region cannot call the `us.*` cross-region profile, with:

```
Research sub-agent LLM error: litellm.BadRequestError: BedrockException -
{"message":"The provided model identifier is invalid."}
```

## Cause

`agent/tools/research_tool.py:225` hard-codes the prefix:

```python
if main_model.startswith("bedrock/") and "anthropic" in main_model:
    return "bedrock/us.anthropic.claude-sonnet-4-6"
```

AWS Bedrock exposes the same Sonnet profile under multiple prefixes (`us.`, `apac.`, `global.`, ...); only `us.` is hard-coded here. The main agent passes the user-chosen profile through to LiteLLM unmodified, so it works — only the sub-agent overrides it.

## Reproduction

1. AWS region without access to `us.*` Anthropic inference profiles (e.g. `ap-northeast-1`).
2. Set the main model to a non-`us.` Bedrock Anthropic profile, e.g. `/model bedrock/global.anthropic.claude-opus-4-7`.
3. Send any prompt that triggers the research tool. The sub-agent fails with the error above; the main agent itself is unaffected.

Confirmed with a direct LiteLLM call from `ap-northeast-1`:

```
bedrock/us.anthropic.claude-sonnet-4-6  -> BedrockException "The provided model identifier is invalid."
bedrock/jp.anthropic.claude-sonnet-4-6  -> OK
```

## Fix

Make the sub-agent inherit the prefix from the main model rather than overriding it. The cheaper sub-agent model name (`claude-sonnet-4-6`) is preserved; only the prefix follows the main agent's choice. PR: #185.

taxfree-python:
Fixes #184.

## Problem

The research sub-agent hard-codes `bedrock/us.anthropic.claude-sonnet-4-6`, which AWS Bedrock rejects with `"The provided model identifier is invalid."` whenever the caller's region/account cannot reach the `us.*` cross-region inference profile (e.g. `ap-northeast-1`). The main agent itself is unaffected because the user-chosen profile id is passed straight through to LiteLLM — only the sub-agent overrides it.

## Fix

Inherit the inference-profile prefix from the main model. If the user is running `bedrock/jp.anthropic.claude-opus-4-7`, the sub-agent now picks `bedrock/jp.anthropic.claude-sonnet-4-6`; same for `global.`, `apac.`, etc. The sub-agent stays on the cheaper Sonnet model — only the prefix follows the main agent's choice, matching the project's existing convention of letting the user's explicit profile selection drive Bedrock routing.

```python
if main_model.startswith("bedrock/") and "anthropic" in main_model:
    head = main_model.split("/", 1)[1]
    region_prefix = head.split(".", 1)[0]
    return f"bedrock/{region_prefix}.anthropic.claude-sonnet-4-6"
```

The existing `us.` test case is preserved (the new logic still produces `us.` for a `us.` main model), and a new test covers `global.` / `apac.` prefixes.

## Verification

Direct LiteLLM call from `ap-northeast-1` before/after, against the same model the function returns:

```
bedrock/us.anthropic.claude-sonnet-4-6  -> BedrockException "The provided model identifier is invalid."  (the reported bug)
bedrock/jp.anthropic.claude-sonnet-4-6  -> OK   (what the patched function emits for the user's main model)
```

Unit tests pass:

```
$ pytest tests/unit/test_cli_rendering.py
======================== 7 passed, 5 warnings in 6.51s =========================
```

## Caveats

This makes the sub-agent work for every prefix under which `claude-sonnet-4-6` is published on Bedrock (currently `us.`, `jp.`, `global.`). If a user is on an exotic prefix that has no `sonnet-4-6` variant, the sub-agent will still fail on that environment — but that's also true of the main agent on that prefix today, and addressing it would expand the scope beyond this fix.

### DIFF
--- a/agent/tools/research_tool.py
+++ b/agent/tools/research_tool.py
@@ -224,7 +224,12 @@ def _get_research_model(main_model: str) -> str:
     if main_model.startswith("anthropic/"):
         return "anthropic/claude-sonnet-4-6"
     if main_model.startswith("bedrock/") and "anthropic" in main_model:
-        return "bedrock/us.anthropic.claude-sonnet-4-6"
+        # Mirror the main model's Bedrock inference-profile prefix so the
+        # user's explicit profile choice carries through to the sub-agent
+        # instead of being silently overridden by a hard-coded one.
+        head = main_model.split("/", 1)[1]
+        region_prefix = head.split(".", 1)[0]
+        return f"bedrock/{region_prefix}.anthropic.claude-sonnet-4-6"
     # For non-Anthropic models (HF router etc.), use the same model
     return main_model
 
--- a/tests/unit/test_cli_rendering.py
+++ b/tests/unit/test_cli_rendering.py
@@ -22,6 +22,17 @@ def test_bedrock_anthropic_research_model_stays_on_bedrock():
     )
 
 
+def test_bedrock_research_model_inherits_main_region_prefix():
+    assert (
+        _get_research_model("bedrock/global.anthropic.claude-opus-4-7")
+        == "bedrock/global.anthropic.claude-sonnet-4-6"
+    )
+    assert (
+        _get_research_model("bedrock/apac.anthropic.claude-sonnet-4-20250514-v1:0")
+        == "bedrock/apac.anthropic.claude-sonnet-4-6"
+    )
+
+
 def test_non_anthropic_research_model_is_unchanged():
     assert _get_research_model("openai/gpt-5.4") == "openai/gpt-5.4"
 


taxfree-python:
Note on the failing `review` check: it appears to be the standard `anthropics/claude-code-action@v1` permission issue affecting external-fork PRs (`Actor does not have write permissions to the repository`). I see the same failure on other recent fork-based PRs (e.g. #180, #181, #182), while internal-branch PRs pass. The change here is unrelated — happy to rebase or adjust if you'd like a fresh run after any workflow tweak.

aryan5v:
@taxfree-python the failed Claude check is expected for external fork PRs. It only runs automatically for contributors with write access to avoid potential abuse of the review workflow and repo tokens. Maintainers can still trigger it

fglogan:
closed per maintainer request

lewtun:

The Bedrock-region research-model fix is superseded by current main's research model resolution.

PR #188, opened 2026-04-29
Add OpenRouter provider support

kenrogers:
## Summary

Adds explicit OpenRouter model support to ML Intern's LiteLLM parameter resolver.

ML Intern already treats bare model ids as Hugging Face Router ids, and this PR
keeps that behavior unchanged. The fix here is narrower: when a user passes an
explicit provider-prefixed model id like `openrouter/...`, route it through
LiteLLM's OpenRouter provider instead of letting it fall through to the Hugging
Face Router fallback.

That avoids surprising endpoint/auth behavior for an explicit third-party
provider namespace while preserving the existing Hugging Face Router default for
bare model ids.

Example:

```bash
ml-intern --model openrouter/anthropic/claude-sonnet-4-5 "your prompt"
```

## Details

- routes `openrouter/...` model ids through LiteLLM's OpenRouter provider before the Hugging Face Router fallback
- keeps the existing Hugging Face Router fallback for bare and unrecognized model ids
- reads `OPENROUTER_API_KEY` and optional `OPENROUTER_API_BASE`
- supports optional OpenRouter attribution headers via `OR_SITE_URL` and `OR_APP_NAME`
- forwards supported `reasoning_effort` values consistently with the existing OpenAI-style resolver behavior
- documents the env var and CLI usage in the README

## Tests

```bash
uv run --extra dev pytest tests/unit/test_llm_params.py
```

Result: `13 passed`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Create a `.env` file in the project root (or export these in your shell):
 ```bash
 ANTHROPIC_API_KEY=<your-anthropic-api-key> # if using anthropic models
 OPENAI_API_KEY=<your-openai-api-key> # if using openai models
+OPENROUTER_API_KEY=<your-openrouter-api-key> # if using openrouter models
 HF_TOKEN=<your-hugging-face-token>
 GITHUB_TOKEN=<github-personal-access-token> 
 ```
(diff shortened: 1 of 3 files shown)


fglogan:
closed per maintainer request

lewtun:

The direct OpenRouter provider path is stale and no longer matches the current HF Router/provider policy.

PR #193, opened 2026-04-29
Auto-share session traces to a private HF dataset

### LINKED ISSUE #191: automatically share my traces on HF
Ideally private by default and I can switch them to public manually with a format compatible to https://huggingface.co/changelog/agent-trace-viewer

DarshanCode2005:
Resolves https://github.com/huggingface/ml-intern/issues/191

## What

Every ml-intern session now gets uploaded to your own private Hugging Face
dataset (default `{hf_user}/ml-intern-traces`) in the Claude Code JSONL
format that the [HF Agent Trace Viewer](https://huggingface.co/changelog/agent-trace-viewer)
auto-detects. The dataset is created private. You can flip it to public
from inside the CLI.

The existing upload to `smolagents/ml-intern-sessions` is unchanged. That
dataset still feeds the backend KPI scheduler.

## Why

Users wanted a way to browse, share, and debug their own runs through the
HF trace viewer without exposing them publicly by default.

### Config

- `agent/config.py`: added `share_traces: bool = True` and
  `personal_trace_repo_template: str = "{hf_user}/ml-intern-traces"`.
- `configs/cli_agent_config.json` and `configs/frontend_agent_config.json`:
  surface the same defaults.

### Uploader

- `agent/core/session_uploader.py`: rewrote the subprocess uploader.
  - New `to_claude_code_jsonl(trajectory)` converter. Maps litellm messages
    to user / assistant / tool_use / tool_result blocks. Deterministic
    SHA-1 UUIDs keyed by `session_id::role::idx` so re-uploads keep the
    parent chain stable. System prompts are skipped.
  - New CLI flags: `--format {row, claude_code}`, `--token-env`,
    `--private`.
  - Per-format status tracking on the local trajectory: `upload_status`
    for the org dataset and `personal_upload_status` for the user repo,
    so a failure on one path does not clobber the other.
  - Token resolution: `--token-env` overrides the org fallback chain
    (`HF_SESSION_UPLOAD_TOKEN`, `HF_TOKEN`, `HF_ADMIN_TOKEN`). The
    personal upload uses `HF_TOKEN` directly.
  - `private` is now passed into `create_repo` instead of being hardcoded.
    Existing repos keep whatever visibility the user set.

### Session wiring

- `agent/core/session.py`:
  - Added `_personal_trace_repo_id()`. Returns `None` when sharing is
    disabled, the user is anonymous, or the template is missing.
  - Added `_spawn_uploader()` helper to keep subprocess args in one place.
  - `save_and_upload_detached` now spawns two detached subprocesses: one
    for the org dataset (row format, existing token chain) and one for
    the per-user private dataset (Claude Code format, `HF_TOKEN`).
  - `retry_failed_uploads_detached` now accepts `personal_repo_id` and
    runs both retry passes in parallel.

- `agent/core/agent_loop.py`: pass
  `personal_repo_id=session._personal_trace_repo_id()` into the startup
  retry so personal uploads that failed in a prior session get
  re-attempted on next launch.

### CLI

- `agent/main.py`: new `/share-traces` slash command.
  - `/share-traces` shows the dataset URL and current visibility.
  - `/share-traces public` flips to public via
    `HfApi.update_repo_settings`.
  - `/share-traces private` flips back to private.
  - Idempotent `create_repo` so the first flip works even before any
    session has been saved.
  - Uses the user's own `HF_TOKEN`.
- `agent/utils/terminal_display.py`: added the command to the help text.

### Docs

- `README.md`: new "Sharing Traces" section covering default behavior,
  the slash command, the opt-out flag, and how to override the template.

## Edge cases handled

- Anonymous user (no resolvable HF username): personal upload is skipped.
- `HF_TOKEN` missing: personal upload is skipped, org upload is unaffected.
- Pre-existing local session logs created before this change: the personal
  retry path skips files that have never been tagged, so we do not
  suddenly re-upload old sessions to a newly created personal repo.
- Heartbeat saves: the same trajectory is rewritten every minute. Both
  upload paths run on every save, the Hub deduplicates by content hash so
  the cost is small.

### DIFF
--- a/agent/core/session_uploader.py
+++ b/agent/core/session_uploader.py
@@ -253,12 +253,18 @@ def to_claude_code_jsonl(trajectory: dict) -> list[dict]:
     return out
 
 
+def _scrub_session_for_upload(data: dict) -> dict:
+    """Best-effort scrub of transcript fields before any upload temp file."""
+    scrubbed = dict(data)
+    scrubbed["messages"] = _scrub(data.get("messages") or [])
+    scrubbed["events"] = _scrub(data.get("events") or [])
+    scrubbed["tools"] = _scrub(data.get("tools") or [])
+    return scrubbed
+
+
 def _write_row_payload(data: dict, tmp_path: str) -> None:
     """Single-row JSONL (existing format) — used by KPI scheduler."""
-    scrubbed_messages = _scrub(data["messages"])
-    scrubbed_events = _scrub(data["events"])
-    scrubbed_tools = _scrub(data.get("tools") or [])
-
+    scrubbed = _scrub_session_for_upload(data)
     session_row = {
         "session_id": data["session_id"],
         "user_id": data.get("user_id"),
@@ -266,9 +272,9 @@ def _write_row_payload(data: dict, tmp_path: str) -> None:
         "session_end_time": data["session_end_time"],
         "model_name": data["model_name"],
         "total_cost_usd": data.get("total_cost_usd"),
-        "messages": json.dumps(scrubbed_messages),
-        "events": json.dumps(scrubbed_events),
-        "tools": json.dumps(scrubbed_tools),
+        "messages": json.dumps(scrubbed["messages"]),
+        "events": json.dumps(scrubbed["events"]),
+        "tools": json.dumps(scrubbed["tools"]),
     }
 
     with open(tmp_path, "w") as tmp:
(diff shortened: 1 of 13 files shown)


agent/core/session_uploader.py
```diff
@@ -46,96 +387,63 @@ def upload_session_as_file(
...
+        hf_token = _resolve_token(token_env)
         if not hf_token:
-            # Update status to failed
-            data["upload_status"] = "failed"
+            data[status_key] = "failed"
             with open(session_file, "w") as f:
                 json.dump(data, f, indent=2)
             return False
 
-        # Scrub secrets (HF tokens, API keys, etc.) from messages + events
```
lewtun, line 68: Why was this removed? We should retain some best-effort scrubbing

lewtun (approved):
Thanks a lot for the nice PR @DarshanCode2005 ! LGTM now with some fixes I pushed

DarshanCode2005:
Thanks @lewtun. Would it be alright if I emailed you for some career advice?

PR #194, opened 2026-04-30
Harden sandbox token handling

lewtun:
## Summary
- Create sandbox Spaces as private Docker Spaces by default while keeping all command/file endpoints authenticated.
- Remove user HF credentials from sandbox environments; only Trackio non-secret variables are allowed.
- Store only a SHA-256 hash of the sandbox control token in the Space and keep the raw token client-side.
- Split private Space routing auth from sandbox app auth: Authorization remains the HF bearer token for hf.space, while X-Sandbox-API-Token authenticates FastAPI control endpoints.
- Redact token-like tool outputs/events before they enter model context, frontend events, or persistence.

## Security posture
- Private Space visibility blocks anonymous platform access before requests reach the sandbox app.
- /api/health stays public for readiness; /api/bash, /api/read, /api/write, /api/edit, /api/exists, and /api/kill require sandbox auth.
- HF_TOKEN and Hugging Face Hub token aliases are rejected from sandbox variable injection and are no longer accepted as sandbox API auth fallbacks.
- The sandbox environment contains SANDBOX_API_TOKEN_SHA256, not the raw control token.

## Tests
- uv run pytest tests/unit/test_sandbox_api_auth.py tests/unit/test_redact.py tests/unit/test_sandbox_already_active_message.py -q
- ML_INTERN_LIVE_SANDBOX_TESTS=1 uv run pytest tests/integration/test_live_sandbox_auth.py -q

## Notes
- Live test created a real private Space, verified unauthenticated access is rejected, exercised authenticated bash/read/write/reconnect, and confirmed no HF_TOKEN, HUGGINGFACE_HUB_TOKEN, or raw SANDBOX_API_TOKEN appears in env output.
- Full unit suite currently has two unrelated pre-existing doom-loop assertion failures: tests expect DOOM LOOP DETECTED while implementation returns [SYSTEM: REPETITION GUARD].

### DIFF
--- a/agent/core/agent_loop.py
+++ b/agent/core/agent_loop.py
@@ -1147,6 +1147,12 @@ class Handlers:
 
                     # 4. Record results and send outputs (order preserved)
                     for tc, tool_name, tool_args, output, success in results:
+                        try:
+                            from agent.core.redact import scrub_string
+
+                            output = scrub_string(output)
+                        except Exception:
+                            pass
                         tool_msg = Message(
                             role="tool",
                             content=output,
--- a/agent/core/redact.py
+++ b/agent/core/redact.py
@@ -16,8 +16,8 @@ from typing import Any
 # Patterns are conservative: they only match tokens with the canonical prefix
 # and a minimum body length so we don't paint over normal text.
 _PATTERNS: list[tuple[re.Pattern, str]] = [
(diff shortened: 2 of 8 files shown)


lewtun:
Closing as this is too strict / breaks core functionality (need token to pull / push repos)

Issue #196, opened 2026-05-01
Optional LangFuse callback for LLM call observability (opt-in)


taxfree-python:
## Context

ml-intern's observability pipeline today writes `Event`s (`llm_call`, `hf_job_*`, `sandbox_*`, …) to a Hugging Face Dataset, with cost/usage extracted from litellm responses (`agent/core/telemetry.py`). PR #105 set this up, PR #179 closed the cost-attribution gap by tagging the 5 LLM call sites (`main` / `research` / `compaction` / `effort_probe` / `restore`).

This is great for the hosted analytics + SFT use case. It's less convenient when:

- **Self-hosting** — operators running ml-intern on their own infra often already have a self-hosted LangFuse / Phoenix / Langsmith and would like LLM traces to land there with no extra plumbing.
- **SaaS LangFuse users** — teams already on cloud.langfuse.com for their other LLM apps want the same single pane of glass for ml-intern.
- **Local development** — seeing one LLM turn's prompts / tool calls / response in a UI immediately, without round-tripping through the HF Dataset → DuckDB pipeline.

## Proposal

Expose litellm's native LangFuse callback as an **opt-in** integration, gated on **all three** environment variables being set — including `LANGFUSE_HOST`, so the destination is always an explicit choice (no silent fallback to litellm's SaaS default).

Concretely, somewhere during agent startup (e.g. `agent/config.py` or `agent/core/agent_loop.py`):

```python
import os, litellm

if (
    os.getenv("LANGFUSE_HOST")           # destination must be explicit — no silent fallback
    and os.getenv("LANGFUSE_PUBLIC_KEY")
    and os.getenv("LANGFUSE_SECRET_KEY")
):
    if "langfuse" not in litellm.success_callback:
        litellm.success_callback.append("langfuse")
    if "langfuse" not in litellm.failure_callback:
        litellm.failure_callback.append("langfuse")
```

Both deployment shapes are supported:

- **Self-host:** `LANGFUSE_HOST=https://langfuse.internal.example.com`
- **SaaS:** `LANGFUSE_HOST=https://cloud.langfuse.com` (or `https://us.cloud.langfuse.com`)

The existing `kind` tag from PR #179 can be forwarded as `metadata={"tags": [kind], "trace_user_id": session.user_id}` at each call site for free filtering in the LangFuse UI.

## Privacy

The litellm → LangFuse callback ships **the full payload** of every LLM call to whatever `LANGFUSE_HOST` points at:

[...]

PR #198, opened 2026-05-01
feat(observability): opt-in LangFuse callback for litellm calls

taxfree-python:
## Summary

Refs #196.

Adds a small, env-gated hook that registers litellm's native LangFuse callback when `LANGFUSE_HOST` + `LANGFUSE_PUBLIC_KEY` + `LANGFUSE_SECRET_KEY` are all set. With any of the three unset the integration is a no-op and runtime behavior is identical to today.

The primary HF-Dataset-based telemetry pipeline (`agent/core/telemetry.py`) is **unchanged** — this is purely an additional side channel for operators who want LLM traces in their own LangFuse instance.

## Why mandatory `LANGFUSE_HOST`

litellm's default LangFuse host is `cloud.langfuse.com`. If the gate were just `PUBLIC_KEY + SECRET_KEY`, an operator who only meant to set keys for a self-hosted instance — but forgot `LANGFUSE_HOST` — would silently exfiltrate prompts to a third-party SaaS. Requiring `LANGFUSE_HOST` makes "where does the data go" a conscious decision. Both deployment shapes are supported:

- **Self-host:** `LANGFUSE_HOST=https://langfuse.internal.example.com`
- **SaaS:** `LANGFUSE_HOST=https://cloud.langfuse.com`

## Scope (intentionally minimal)

- New `agent/core/observability.py` with `setup_langfuse()` (env-gated, idempotent)
- One-line call from `agent/config.py:load_config()` — covers both CLI (`agent/main.py`) and backend (`backend/session_manager.py` module-init) without a separate hook
- `pyproject.toml`: new `[observability]` optional-dep group; `langfuse` is **not** a hard dep
- `agent/README.md`: new "Observability (optional)" section
- `tests/unit/test_observability.py`: gate-passes / gate-blocks / idempotent

**Out of scope (deferred to a follow-up if there's interest):** forwarding the existing `kind` tag from PR #179, plus `session.user_id` / `session.session_id`, as `metadata={...}` to each litellm call. That's a mechanical change touching the 7 acompletion call sites and is much easier to review on its own.

## Test plan

- [x] `pytest tests/unit/test_observability.py -v` — 5 cases pass (1 gate-passes, 3 parametrized gate-blocks, 1 idempotent)
- [x] `pytest tests/unit/test_config.py` — existing config tests still pass (no `load_config` semantics change when env vars unset)
- [ ] Maintainer manual: with `LANGFUSE_*` unset, run the agent — no log lines about LangFuse, no errors
- [ ] Maintainer manual: with vars set + a LangFuse instance reachable, run the agent — "LangFuse observability enabled" log line + traces visible in the LangFuse UI

## Note on CI

The `claude-code-action` review workflow fails on every external-fork PR (`pull_request` triggers don't grant write tokens to fork actors — confirmed across recent fork PRs). The red `review` check is expected and not contributor-fixable; the change itself does not alter any workflow.

### DIFF
--- a/agent/README.md
+++ b/agent/README.md
@@ -19,3 +19,37 @@ Async agent loop with LiteLLM.
 | **`context_manager/`** | Manages conversation history, very rudimentary context engineering support | Implement intelligent context engineering to keep the agent on track |
 | **`config.py`** | Loads JSON config for the agent | Support different configs etc. |
 | **`main.py`** | Interactive CLI with async queue architecture (submission→agent, agent→events) (simple way to interact with the agent now)| Serve as reference implementation for other UIs (web, API, programmatic) |
+
+## Observability (optional)
+
+LLM calls can additionally be streamed to a [LangFuse](https://langfuse.com)
+instance — useful for local development and for self-hosted deployments
+that already run LangFuse / Phoenix / Langsmith. The primary
+HF-Dataset-based telemetry pipeline (`agent/core/telemetry.py`) is unchanged.
+
+Set the LangFuse host plus both keys to opt in. Either env-var name for
+the host works — Langfuse SDK v4 issues credentials as `LANGFUSE_BASE_URL`,
+while litellm's callback reads `LANGFUSE_HOST`; this integration accepts
+either and mirrors the value through to litellm:
+
+```
+LANGFUSE_BASE_URL=https://your-langfuse.example.com   # or LANGFUSE_HOST=...
+LANGFUSE_PUBLIC_KEY=pk-...
+LANGFUSE_SECRET_KEY=sk-...
+```
+
+Both self-hosted LangFuse and the SaaS endpoint
+(`https://cloud.langfuse.com`) are supported. The host is mandatory so the
+destination is always an explicit choice — there is no silent fallback.
+With any of the three vars unset the integration is a no-op.
+
+Install the optional dependency:
+
+```
+pip install ml-intern[observability]
+```
+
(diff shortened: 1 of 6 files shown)


taxfree-python:
Force-pushed an update with two changes after end-to-end testing against a self-hosted Langfuse instance:

1. **Switched callback from `langfuse` to `langfuse_otel`**. The legacy `langfuse` integration in current litellm releases breaks against Langfuse SDK v4 (`AttributeError: module 'langfuse' has no attribute 'version'`). The OTEL path works against both v3 and v4 and is what Langfuse v4 docs recommend.

2. **Accept `LANGFUSE_BASE_URL` as an alias for `LANGFUSE_HOST`**. Langfuse SDK v4's setup docs issue credentials as `LANGFUSE_BASE_URL`, but litellm's callback only reads `LANGFUSE_HOST` — without this alias the gate would silently no-op for anyone copy-pasting from Langfuse's own docs. We accept either and mirror `BASE_URL` into `HOST` so litellm sees it. `HOST` takes precedence if both are set.

Verified end-to-end with `litellm.completion(...)` + `langfuse_otel`: trace lands in the configured Langfuse instance, retrievable via `/api/public/traces`. Tests cover both env-var names plus precedence; all 7 observability tests pass.

fglogan:
closed per maintainer request

lewtun:

The branch is conflicted and the touched observability/routing files have moved significantly since it was opened.

PR #205, opened 2026-05-01
fix: async plan isolation, blocking API call, and CORS for production

angziii:
## Summary

Three targeted fixes:

- **plan_tool.py: Per-session plan isolation** — The module-level `_current_plan` list was shared across all sessions, causing concurrent sessions to overwrite each other's plans. Replaced with a per-session dict keyed by `session_id`. Updated `get_current_plan()`, `format_plan_display()`, and `print_plan()` to be session-aware.

- **hf_repo_git_tool.py: Async wrapper for `get_repo_discussions`** — Every other `HfApi` call in this file uses `_async_call()` to run in a thread pool, but `get_repo_discussions` was called synchronously, blocking the event loop during the HTTP request. Wrapped it in `_async_call()` with the list materialized inside the thread.

- **backend/main.py: CORS origins for production** — `allow_origins` was hardcoded to localhost-only, causing CORS middleware to reject browser requests when deployed to HF Spaces. Added `SPACE_HOST` env var to the allow list dynamically.

## Test plan
- [x] Code review of diff
- [ ] Verify plan isolation: two concurrent sessions should have independent plans
- [ ] Verify `get_repo_discussions` no longer blocks the event loop
- [ ] Verify CORS allows production HF Spaces URL when `SPACE_HOST` is set

### DIFF
--- a/agent/main.py
+++ b/agent/main.py
@@ -347,8 +347,8 @@ async def event_listener(
                 shimmer.stop()
                 stream_buf.discard()
                 print_turn_complete()
-                print_plan()
                 session = session_holder[0] if session_holder else None
+                print_plan(session=session)
                 if session is not None:
                     await session.send_deferred_turn_complete_notification(event)
                 turn_complete_event.set()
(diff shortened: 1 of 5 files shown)


fglogan:
closed per maintainer request

lewtun:

The branch is conflicted, has had no recent updates, and the thread already indicates closure was requested by a maintainer.

PR #208, opened 2026-05-01
Fallback to free model for gated defaults

lewtun:
## Summary
- keep Claude as the configured default for users who pass the premium model gate
- fall back to Kimi K2.6 when an implicit default session would otherwise fail the gate
- keep explicit Claude/GPT-5.5 selections gated for non-eligible users

## Test
- uv run --extra dev pytest tests/unit/test_agent_model_gating.py

### DIFF
--- a/backend/models.py
+++ b/backend/models.py
@@ -66,6 +66,7 @@ class SessionResponse(BaseModel):
 
     session_id: str
     ready: bool = True
+    model: str | None = None
 
 
 class PendingApprovalTool(BaseModel):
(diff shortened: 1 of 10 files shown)


lewtun:
Addressed in 19d347a by adding coverage for the HF-member explicit gated-model request branch in tests/unit/test_agent_model_gating.py.

Tested with:

```bash
uv run --extra dev pytest tests/unit/test_agent_model_gating.py
```

AbijahKaj:
Hi @lewtun  ! 
Seems like the hosted version became unusable on my side after this being merged, the model is gated to the free ones and I'm not getting any response to my messages from the llm and no error shown either, I don't know if I've expunged my quota or it's just because I deleted the ml-intern-sessions dataset.

I can see in the responses from requests in my browser I still have quota but I'm not getting any response from the llm back.

PR #209, opened 2026-05-01
Stop attaching timestamps to LLM messages

lewtun:
## Summary

- stop attaching local `timestamp` metadata to `litellm.Message` objects
- remove the extra provider-message conversion layer from the earlier version of this PR
- keep the existing dangling-tool-call behavior unchanged

## Testing

- `uv run --extra dev pytest tests/unit/test_dangling_tool_calls.py`
- `uv run --extra dev pytest tests/unit/test_dangling_tool_calls.py tests/unit/test_session_uploader.py tests/unit/test_compaction_loop_break.py`

## Notes

The fix is intentionally small: new messages no longer carry the field that OpenAI-compatible providers rejected.


fglogan:
closed per maintainer request

PR #210, opened 2026-05-02
Retry sandbox hardware requests

lewtun:
Retry transient Hub errors when requesting hardware for a newly duplicated private sandbox Space. This covers the deployed OAuth path where duplicate succeeds but /hardware can briefly return 401/Repository Not Found before the private Space is manageable.


fglogan:
closed per maintainer request

PR #213, opened 2026-05-03
fix(compaction): break infinite loop + truncate oversized messages

jagwar:
## Why

Sessions stuck in a compaction retry loop are silently burning Bedrock budget while staying invisible in the session dataset. Pod logs from 2026-05-03 showed the pattern firing minute-after-minute on multiple replicas:

```
12:14:15  Context compacted: 200001 -> 215566 tokens
12:14:21  Context compacted: 215566 -> 215572 tokens
12:14:22  ContextWindowExceededError — forcing compaction
12:14:26  Context compacted: 200001 -> 215544 tokens
12:14:30  Context compacted: 215544 -> 215545 tokens
... (continues for 5+ minutes straight, on multiple pods)
```

**Root cause:** a single message in the "untouched" tail (typically a tool output of 80k+ tokens — `bash` dump, file content, CSV) keeps the post-compact context above the 90% threshold. Compaction triggers again, calls Bedrock again, gets the same 215k result, loops. **~\$3 per Opus retry, indefinitely.** Sessions never reach `_run_session.finally` → `save_and_upload_detached` never fires → cost is real but never lands in `total_cost_usd`.

**Quantitative cross-check:**
- Dataset Bedrock-only coverage on 2026-05-01 = 43% of Cost Explorer (\$6,299 vs \$14,732)
- Bedrock Invocation Logs on 2026-05-02: 11,996 InvokeModel calls vs **5,558 Bedrock events** in the dataset for the same day — half the calls missing, consistent with stuck sessions never uploading.

## What

Three fixes in one PR (single failure mode, single PR):

### 1. `ContextManager._truncate_oversized()` — root cause fix

Before summarization, replace the content of any preserved message (first user, untouched tail) over 50k tokens with a placeholder:

```
[truncated for compaction — original was 87,432 tokens, removed to keep context under 162,000 tokens]
```

This is the only place where individual messages can defeat compaction by being intrinsically too large. Messages in the middle (`messages_to_summarize`) are folded into the summary so their size doesn't matter on its own.

### 2. `ContextManager.compact()` — fail-fast verification

After the summarize call, recompute `running_context_usage` via `litellm.token_counter`. If it's still over the threshold, raise `CompactionFailedError`. The retry-as-circuit-breaker behavior is replaced with fail-fast: better to end the session cleanly than loop on the same useless API call.

### 3. `agent_loop._compact_and_notify()` — graceful session termination

Catch `CompactionFailedError`, emit a `session_terminated` event with `reason=compaction_failed` (dataset records WHY the session ended + the cost incurred up to that point), set `session.is_running=False` to exit the loop. `_run_session.finally` then fires `save_trajectory` normally and the session ends up in the dataset.

## Expected impact

- **Cost reduction:** sessions that previously looped invisibly will now end at first compaction failure. Sample: one stuck session on Opus = ~\$3/iteration × 100s of iterations = \$50-300 lost per stuck session. If even 5-10 sessions/day get stuck, this is \$500-3000/day saved.
- **Dataset coverage:** the missing ~57% of Bedrock spend on the dataset side should significantly close. Sessions that end via `CompactionFailedError` will appear with their cost properly attributed (not as much as a successful session, but at least visible).
- **UX:** users get a clear "your conversation got too large, start a new one" message instead of an unresponsive agent.

## Test plan

- [ ] Read-through to confirm no regression on the success path (compaction works as before when context can fit after summarization).
- [ ] Manual: construct a session with a fake 100k-token tool output in the untouched tail, trigger compaction, verify (a) the message gets truncated, (b) compaction succeeds, (c) session continues normally.
- [ ] Manual: construct a session with system_prompt + first_user > threshold (untruncatable), trigger compaction, verify CompactionFailedError fires, session_terminated event is emitted, `save_trajectory` runs.
- [ ] After deploy: monitor pod logs on prod-114 for the absence of the looping pattern. The signal: `Compaction ineffective` at ERROR level → followed by clean session termination, NOT by another compaction call 3s later.

## Out of scope

- Compaction events on HF Router models (DeepSeek, Kimi, GLM) still get `cost_usd=0` because litellm doesn't have their pricing tables. Separate issue, smaller impact (those are HF Router billing, not Bedrock).
- Deeper rework of how `untouched_messages=5` is computed adaptively. The truncation is sufficient for now; smarter logic can come later.

cc @AkselReedi

### DIFF
--- a/agent/context_manager/manager.py
+++ b/agent/context_manager/manager.py
@@ -508,15 +508,15 @@ class ContextManager:
         idx = len(self.items) - self.untouched_messages
         while idx > 1 and self.items[idx].role != "user":
             idx -= 1
-        # The while loop's `idx > 1` guard expresses "do not include the
-        # system message in recent_messages". But when len(self.items) ==
-        # untouched_messages (the canonical 5-message early-compaction case
-        # — system + user-task + giant-tool-output + user-followup +
-        # assistant-reply), idx initialises to 0 and the while is a no-op.
-        # Without this clamp, recent_messages = self.items[0:] would include
-        # the system message, and the rebuild path duplicates it.
-        if idx < 1:
-            idx = 1
+        # The real invariant is "idx must be strictly after first_user_idx,
+        # otherwise recent_messages overlaps with the messages we put in
+        # head". The walk-back's `idx > 1` guard is necessary (no system in
+        # recent) but insufficient (first_user is also in head and would be
+        # duplicated). Anthropic API rejects two consecutive user messages
+        # with a 400 — bot review on PR #213 caught this on the second clamp
+        # iteration.
+        if idx <= first_user_idx:
+            idx = first_user_idx + 1
 
         recent_messages = self.items[idx:]
         messages_to_summarize = self.items[first_user_idx + 1:idx]
--- a/agent/context_manager/manager.py
+++ b/agent/context_manager/manager.py
@@ -79,6 +79,23 @@ _COMPACT_PROMPT = (
     "will be have to be filled in."
 )
 
+# Per-message ceiling. If a single message in the "untouched" tail is larger
+# than this, compaction can't recover even after summarizing the middle —
+# producing the infinite compaction loop seen 2026-05-03 in pod logs (200k
+# context shrinks to 200k+ because one tool output is 80k tokens). We replace
+# such messages with a placeholder before compaction runs.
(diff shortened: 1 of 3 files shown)


fglogan:
closed per maintainer request

lewtun (approved):
Thanks for the fix! LGTM

PR #219, opened 2026-05-04
Log exception for HF whoami username resolution failure

Goodnight77:
## summary
  - _get_hf_user in agent/main.py swallowed all exceptions silently
  - added module logger and logger.exception call so the traceback is recorded before the function falls back to None


lewtun:

PR #220, opened 2026-05-04
Update README with new model version and usage instructions

Goodnight77:
## summary                                                                                                                                                                                                                       
  - bumped Claude example in README to anthropic/claude-opus-4-7 so it matches the current default model    
  - annotated both model examples with the env var they need  (ANTHROPIC_API_KEY / OPENAI_API_KEY)
  - added a one-liner pointing users at /model for the full suggested list


lewtun (approved):
Thanks for the fix!

PR #227, opened 2026-05-05
Force OAuth refresh after scope changes

lewtun:
## Summary

- force Space users with old OAuth cookies to log in again after the required scope set changes
- centralize the backend's required OAuth scopes and include `write-collections` in that contract
- validate returned OAuth scopes when the provider includes them, and add coverage for stale scope-cookie handling

## Why

The Space README and backend now request `write-collections`, but existing `hf_access_token` cookies can remain valid for up to seven days. Those stale tokens were minted before the collection scope existed, so local testing with a broad PAT worked while the deployed Space still failed to create per-session Hub collections.

## Tests

- `uv run --extra dev pytest tests/unit`
- `uv run ruff check .`

### DIFF
--- a/backend/dependencies.py
+++ b/backend/dependencies.py
@@ -7,6 +7,8 @@
 import logging
 import os
 import time
+from collections.abc import Iterable
+from hashlib import sha256
 from typing import Any
 
 import httpx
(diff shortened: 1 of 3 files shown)


lewtun:
Addressed the review items in 10104d6: added debug visibility for absent OAuth scope fields, documented the Bearer-token freshness asymmetry, and added the missing absent-scope test.\n\nChecks run:\n- uv run --extra dev pytest tests/unit\n- uv run ruff check .\n- uv run ruff format --check .

Issue #94, opened 2026-04-24
Local Model Support


haseebsultankhan:
Kindly add vllm or ollama for local model support if possible? 

Or can someone guide me how to configure that?

Prajwalkumar2005:
1: Ollama
* Install Ollama → `curl -fsSL https://ollama.com/install.sh | sh`
* Pull a model → `ollama pull llama3`
* Run locally → `ollama run llama3`
* Access via API at `http://localhost:11434`

2: vllm
* Install  - pip install vllm
* Serve model  `python -m vllm.entrypoints.openai.api_server --model mistralai/Mistral-7B-Instruct-v0.1`
 . Use OpenAI-compatible API at `http://localhost:8000`

let me know if it works  i took refrence from this in past

haseebsultankhan:
I have already configured myself , will share the fork here , 

I have included Ollama , Ollama Cloud , Google Gemini via api

harshitgavita-07:
Hi @akseljoonas , @abidlabs, @henribonamy and other maintainers 👋

First of all, thanks a lot for releasing ml-intern and for keeping it open-source. It looks amazing as an autonomous ML engineer.

I would really love to use ml-intern with local / free models instead of paid hosted APIs, ideally running everything on my own hardware (e.g. via vLLM, Ollama, or another local OpenAI-compatible endpoint).

Could you please share your thoughts on:

Official support or roadmap for local model backends (vLLM, Ollama, TGI, etc.).

A recommended way to point ml-intern to a self-hosted LLM endpoint instead of a remote provider.

Any configuration examples for running ml-intern fully locally (for people who want to avoid paid/proprietary APIs).

I’m happy to help test configurations, especially for local setups, and provide feedback.

Thanks again for your work on this project!

lhhmmiii:
@haseebsultankhan I have submitted a PR that adds support for Ollama. I hope this helps! 
The PR: #212

fglogan:
closed per maintainer request

lewtun:
Thanks for contributing to ML Intern. The project, including its CLI, has been retired and this repository is being archived. We will therefore not be able to review or merge this work. For continued AI assistance, please use [HuggingChat](https://huggingface.co/chat/).

PR #44, opened 2026-04-21
Add local model support via Ollama

lmassaron:
## Summary
This PR adds support for running the agent with local models via Ollama. It enables users to use the intern without incurring API costs or requiring external API keys for inference.
## Key Changes
- **Ollama Integration**: Integrated local model support using Ollama's OpenAI-compatible endpoint for improved
tool-calling reliability.
 - **Readiness Checks**: The agent now automatically verifies if `ollama serve` is running and if the requested model is
available locally.
- **Interactive Model Management**:
- Added `/add-model <id>` command to dynamically add new local models during a session.
- Added an interactive prompt to pull missing models directly from the CLI.
- **Optional Authentication**: Modified initialization to make the `HF_TOKEN` optional when using local models (prefixed
with `ollama/`).
- **Robust Initialization**: Improved the startup loop to handle local environment failures without hanging.

## How to test
1. Ensure Ollama is running (`ollama serve`).
2. Run the agent: `ml-intern`.
3. Switch to a local model: `/model ollama/qwen3.6` (or use `/add-model`).

PR #228, opened 2026-05-05
Add CLI local model support

lewtun:
## Summary

Adds CLI-only support for local OpenAI-compatible model servers through provider-specific model IDs:

- `ollama/<model>`
- `vllm/<model>`
- `lm_studio/<model>`
- `llamacpp/<model>`

This keeps the web/backend model catalog unchanged and intentionally does not add the generic `openai-compat/` escape hatch yet.

## Background / lineage

This PR takes the narrow CLI routing path from the local-model discussion and builds on ideas from prior PRs:

- References #44 for the Ollama-first CLI support and HF-token startup behavior.
- References #68 for the small LiteLLM OpenAI-compatible routing shape.
- References #166 for local-model feature-gating concerns and strict reasoning-effort handling.

It addresses the CLI slice of #94 without adding frontend/API local model selection.

## Usage examples

Start the local inference server first, then select it with a provider-specific prefix:

```bash
ml-intern --model ollama/llama3.1:8b "summarize this repo"
ml-intern --model vllm/meta-llama/Llama-3.1-8B-Instruct "inspect the training script"
```

In an interactive session:

```text
/model ollama/llama3.1:8b
/model lm_studio/google/gemma-3-4b
/model llamacpp/llama-3.1-8b-instruct
```

Use one shared local endpoint for any local prefix:

```bash
LOCAL_LLM_BASE_URL=http://localhost:8000
LOCAL_LLM_API_KEY=optional-local-key
ml-intern --model vllm/custom-model "inspect the training script"
```

Or override a specific provider when running multiple local servers:

```bash
OLLAMA_BASE_URL=http://localhost:11434
VLLM_BASE_URL=http://localhost:8000
LMSTUDIO_BASE_URL=http://127.0.0.1:1234
LLAMACPP_BASE_URL=http://localhost:8080
```

Base URLs may include or omit `/v1`. Local API keys are optional; when unset, local backends get a non-empty placeholder key by default.

Local endpoint precedence:

```text
base URL:  <PROVIDER>_BASE_URL -> LOCAL_LLM_BASE_URL -> provider default
api key:   <PROVIDER>_API_KEY  -> LOCAL_LLM_API_KEY  -> placeholder
```

`OPENAI_BASE_URL` is intentionally left for `openai/...` models only. For example:

```bash
OPENAI_BASE_URL=http://localhost:8000/v1 ml-intern --model openai/my-local-model
```

## Implementation notes

- Local model IDs route through LiteLLM as `openai/<model>` with the configured provider base URL.
- Local providers do not receive `reasoning_effort`; strict mode raises `UnsupportedEffortError`, while normal runtime drops it.
- `/model` skips HF Router catalog lookup for local IDs and performs a cheap no-effort probe before switching.
- CLI startup no longer requires an HF token when the selected model is local. HF-backed tools still need an HF token to act on the Hub.
- Provider-specific local env vars take precedence over shared `LOCAL_LLM_*` vars, and `OPENAI_BASE_URL` is not reused for local prefixes.

## Validation

```bash
uv run pytest tests/unit/test_llm_params.py tests/unit/test_cli_local_models.py
uv run pytest tests/unit
uv run ruff check .
uv run ruff format --check .
```


lewtun:
Addressed the review feedback from https://github.com/huggingface/ml-intern/pull/228#issuecomment-4381206612 in `925c29e`:

- Clarified `probe_and_switch_model()` docs so transient yellow-warning behavior is documented as non-local only; local probe failures reject the switch.
- Added async coverage for local probe failure preserving the current model and leaving `model_effective_effort` untouched.
- Simplified `_resolve_local_model_params()` validation to remove the redundant triple check.

Validation:

```bash
uv run pytest tests/unit/test_llm_params.py tests/unit/test_cli_local_models.py
uv run pytest tests/unit
uv run ruff check .
uv run ruff format --check .
```

Issue #223, opened 2026-05-04
ML Intern backlog prioritization report - 2026-05-04


lewtun:
# ML Intern Backlog Prioritization

Generated: 2026-05-04T13:18:56.926357+00:00
Model: `anthropic/claude-opus-4-6`

Sources: github_issue=24, github_pr=50, hf_discussion=14

## Summary

Critical security PRs (#96, #85, #83) and agent-quality fix (#88) are reviewed and ready to merge. Hosted Space has multiple reliability regressions (session zombies, Pro detection, onboarding CSS, blank pages) blocking paying users. Provider expansion (local models, Bedrock, Azure, Gemini) is the top feature theme with 6+ PRs. Cost guardrails and session durability are strategic priorities.

## Can Be Closed

No high-confidence resolved-in-main candidates found.

## Highest Impact Next

1. **Merge 3 security PRs: CVE patch, sandbox auth, SSRF fix** (impact 5/5, effort 1/5, confidence 0.95)
   - Recommendation: Merge PR #96 (authlib CVE), #85 (sandbox bearer auth), #83 (SSRF origin validation) — all reviewed APPROVE/safe-to-merge, effort 1 each
   - Rationale: Unauthenticated sandbox RCE + HF_TOKEN exfil, SSRF token leak, and CRITICAL CVE — all exploitable today with reviewed fixes waiting
   - Next action: Rebase #85 to resolve dirty state; approve and merge all three in one batch
   - Sources: [github_pr#96](https://github.com/huggingface/ml-intern/pull/96), [github_pr#85](https://github.com/huggingface/ml-intern/pull/85), [github_pr#83](https://github.com/huggingface/ml-intern/pull/83)

2. **Merge thinking_blocks fix (PR #88)** (impact 5/5, effort 1/5, confidence 0.95)
   - Recommendation: Merge PR #88 — reviewed 'SAFE TO MERGE, LOW RISK', 51 lines, 1 file. Fixes silent quality degradation on every Anthropic session.
   - Rationale: Extended thinking silently drops after first tool call on all Anthropic models — degrades agent quality for the entire session
   - Next action: Confirm rebase is clean and merge
   - Sources: [github_pr#88](https://github.com/huggingface/ml-intern/pull/88)

3. **Fix onboarding CSS overflow blocking Start button** (impact 5/5, effort 1/5, confidence 0.95)
[...]

Issue #197, opened 2026-05-01
feat: support OpenRouter and custom OpenAI-compatible base URLs


sowhan:
## Problem
ml-intern currently only works with Anthropic and OpenAI directly. 
There's no way to point the OpenAI client at a custom base URL, 
which means OpenRouter (and other compatible providers like Together, 
Groq, etc.) aren't supported.

## Proposed solution
Respect `OPENAI_BASE_URL` env var when initializing the OpenAI client — 
same pattern used by Claude Code, smolagents' `OpenAIModel`, and the 
broader OpenAI SDK ecosystem.

Optionally, a `--base-url` CLI flag for explicit overrides.

## Example usage
```bash
export OPENAI_API_KEY="sk-or-v1-..."
export OPENAI_BASE_URL="https://openrouter.ai/api/v1"
ml-intern --model openai/deepseek/deepseek-r1 "fine-tune a Whisper model"
```

This unlocks 300+ models on OpenRouter without any new dependencies.

## Question
Would you accept a PR for this? Happy to keep it minimal — just the 
base URL passthrough, no refactoring.

fglogan:
closed per maintainer request

haocn-ops:
If you add `OPENAI_BASE_URL`, I’d regression-test two paths separately: initial client construction and any retry/fallback path. A common bug is that the first request uses the custom endpoint, while fallback reconstructs a default OpenAI client and sends traffic back to `api.openai.com`.

Logging the resolved provider/base URL with the model alias, redacted of course, makes those failures much easier to diagnose.

xianzuyang9-blip:
If you add OPENAI_BASE_URL support, I’d include a small compatibility check before merging: /v1/models discovery, model alias handling, non-streaming chat, streaming chat, tool-call/tool-result behavior, and provider error messages. OpenRouter, LiteLLM, local vLLM, and smaller proxy endpoints all differ a little even when they claim OpenAI compatibility.

I have a small GPT/Claude-compatible /v1 endpoint that can be used as a low-volume control endpoint for testing this kind of change: https://99887123.xyz/v1

HerShin5:
+1 for keeping this generic rather than OpenRouter-specific.

I'm testing KiroForge as another OpenAI-compatible gateway (`https://kiroforge.cloud/v1`), and the same `OPENAI_BASE_URL` + API key + model pattern should work without extra dependencies if the OpenAI client constructor passes through `baseURL`.

[...]

PR #232, opened 2026-05-06
Add native OpenRouter model support

lewtun:
## Summary
- add native `openrouter/...` model-id handling before the Hugging Face Router fallback
- keep `openai/...` reserved for direct OpenAI usage and intentionally do not wire `OPENAI_BASE_URL`
- update CLI model-switching help/docs and ensure OpenRouter ids bypass HF Router catalog lookup

## Tests
- `uv run --extra dev pytest tests/unit/test_llm_params.py tests/unit/test_cli_rendering.py`
- `uv run ruff check .`

Refs #223, #197, #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Create a `.env` file in the project root (or export these in your shell):
 ```bash
 ANTHROPIC_API_KEY=<your-anthropic-api-key> # if using anthropic models
 OPENAI_API_KEY=<your-openai-api-key> # if using openai models
+OPENROUTER_API_KEY=<your-openrouter-api-key> # if using openrouter models
 HF_TOKEN=<your-hugging-face-token>
 GITHUB_TOKEN=<github-personal-access-token> 
 ```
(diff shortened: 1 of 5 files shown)


husayni:
Drive-by — same approach landed in my local fork while testing OpenRouter+Kimi K2.6 yesterday, so I dug into the Claude-bot review and have concrete diffs for the three P1s if it helps move this along.

**P1.1 — auth-error hint** (`agent/core/agent_loop.py:518-520`)
```diff
         "  • Anthropic:   export ANTHROPIC_API_KEY=sk-...\n"
         "  • OpenAI:      export OPENAI_API_KEY=sk-...\n"
+        "  • OpenRouter:  export OPENROUTER_API_KEY=sk-or-...\n"
         "  • HF Router:   export HF_TOKEN=hf_...\n\n"
```

**P1.2 — research sub-agent step-down** (`agent/tools/research_tool.py:222-226`)
```diff
     if main_model.startswith("anthropic/"):
         return "anthropic/claude-sonnet-4-6"
     if main_model.startswith("bedrock/") and "anthropic" in main_model:
         return "bedrock/us.anthropic.claude-sonnet-4-6"
+    if main_model.startswith("openrouter/anthropic/"):
+        return "openrouter/anthropic/claude-sonnet-4-6"
     return main_model
```
Without this, anyone routing Opus 4.7 through OpenRouter pays Opus rates for every research sub-agent — material cost regression vs. the `anthropic/...` path.

**P1.3 — `test_openai_base_url_is_not_forwarded` doesn't enforce the README claim**
The test only checks the params dict. LiteLLM's OpenAI adapter honors `OPENAI_BASE_URL` from env regardless of params, so a user with `OPENAI_BASE_URL=https://openrouter.ai/api/v1` already set *will* have their `openai/...` calls silently re-routed — the invariant the README promises isn't actually there. Two options:
- **(a)** Tighten the test: monkeypatch `litellm.acompletion` and assert the resolved `api_base` for `openai/gpt-5.5` doesn't equal `OPENAI_BASE_URL` when the env var is set.
- **(b)** Soften the README: "we don't pass `api_base`; LiteLLM still honors `OPENAI_BASE_URL` from your environment — set it deliberately."

Happy to push these as commits on your branch if you'd rather not juggle them, or open a follow-up PR after this merges. Also noticed the branch is `CONFLICTING` against main — happy to rebase if useful.

lewtun:
Closing in favour of routing everything via HF inference providers

PR #233, opened 2026-05-06
Implement `/resume` command for CLI

cmpatino:
Adds a CLI `/resume` slash command for continuing from local session logs in `session_logs/`.

The feature implementation includes the following:
- Adds an interactive newest-first session-log picker, plus direct lookup by index, path, filename, or session id prefix.
- Routes resume through the agent submission loop with `OpType.RESUME`, matching `/undo` and `/compact`.
- Restores saved non-system messages while keeping the current system prompt, updates the model, recomputes context usage, and clears stale pending approvals.
- Handles identity safely: matching-user logs continue in-place and reuse the saved log path; cross-user/anonymous mismatches fork into the current session.
- Resets restored event history to a `resumed_from` marker to avoid cost double-counting.
- Warns when restored messages contain redacted token placeholders.

### DIFF
--- a/agent/core/agent_loop.py
+++ b/agent/core/agent_loop.py
@@ -7,6 +7,7 @@ import json
 import logging
 import time
 from dataclasses import dataclass, field
+from pathlib import Path
 from typing import Any
 
 from litellm import (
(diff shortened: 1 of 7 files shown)


lewtun (approved):
Thanks for the nice feature - LGTM!

PR #235, opened 2026-05-07
feat: support OpenRouter and custom OpenAI-compatible base URLs (#197)

aadithya12ctrl:


lewtun:

The branch is conflicted and the related OpenRouter/custom-base-url work now points toward a generic OPENAI_BASE_URL-style approach rather than this prefix-specific patch.

PR #237, opened 2026-05-07
Add CLI sandbox runtime and fix HF Jobs script paths

lewtun:
## Summary
- Add an opt-in CLI sandbox tool runtime so local models can use HF Space sandbox tools, including `sandbox_create`.
- Keep the default CLI runtime unchanged: `bash`, `read`, `write`, and `edit` operate on the local filesystem unless sandbox tools are explicitly enabled.
- Treat bare Python filenames such as `train_smollm2.py` as sandbox script paths when resolving `hf_jobs` payloads.
- Clarify prompt/tool guidance so `hf_jobs.script` uses inline code, sandbox files, or public/raw URLs, not host-local paths like `/fsx/...`.

## Why
This came up while testing local model workflows from #228. Small local models can create a training script, then call `hf_jobs` with `script: "train_smollm2.py"` or a host-local path. Previously bare filenames were not resolved from the sandbox, so the remote HF Job could fail with `No such file or directory`.

The same testing also showed that CLI local mode intentionally swaps sandbox tools for local filesystem tools, which means `sandbox_create` is unavailable by default. This PR makes the runtime explicit and opt-in instead of changing the default behavior.

## Usage
Default local filesystem tools:

```bash
ml-intern --model llamacpp/ggml-org/gemma-3-1b-it-GGUF
```

Opt into HF Space sandbox tools:

```bash
ml-intern --model llamacpp/ggml-org/gemma-3-1b-it-GGUF --sandbox-tools
```

Or via config:

```json
{
  "tool_runtime": "sandbox"
}
```

Sandbox mode requires an HF token even when the LLM itself is local, because it creates private HF Spaces.

## Behavior
- `tool_runtime: "local"` uses local `bash/read/write/edit` and excludes `sandbox_create`.
- `tool_runtime: "sandbox"` uses sandbox `bash/read/write/edit` and includes `sandbox_create`.
- Sandbox CLI sessions preload a `cpu-basic` sandbox and clean it up on shutdown.
- Bare sandbox Python files like `train_smollm2.py`, `./train_smollm2.py`, and `/app/train_smollm2.py` are read and submitted to HF Jobs as script content.
- URLs such as `https://example.com/train.py` are still passed through as Jobs URLs.
- Host-local paths such as `/Users/...`, `/home/...`, and `/fsx/...` are explicitly discouraged in the system prompt.

## Example Failure Fixed
Before:

```json
{"script": "train_smollm2.py"}
```

could fail remotely as:

```text
Failed to spawn: train_smollm2.py
Caused by: No such file or directory
```

After this change, the CLI reads `train_smollm2.py` from the sandbox and submits the script content, matching the behavior for explicit paths like `./train_smollm2.py`.

## Tests
- `uv run pytest tests/unit`
- `uv run ruff check .`

Follow-up to #228.


lewtun:
Addressed in `4be6975`: updated the `hf_jobs` `script` parameter description to mention bare sandbox Python filenames like `train.py`, alongside absolute/relative paths and URLs. I also added a unit test that pins the model-facing description.

Checks:
- `uv run pytest tests/unit`
- `uv run ruff check .`

PR #241, opened 2026-05-07
Skip sandbox spaces for artifact metadata

lewtun:
## Summary

- Skip ML Intern sandbox Spaces (`sandbox-<8 hex>`) before artifact registration attempts any repo-card upload or collection update.
- Apply the same guard inside the generated `sitecustomize` hook used by sandbox and HF Jobs Python processes.
- Add regression coverage for both the direct registration path and generated hook path.

## Why

Adding the `ml-intern` tag updates a Space README, which restarts the Space. Sandbox Spaces are ephemeral execution environments, so tagging them can interrupt the running sandbox. These repos should not be treated as generated user-facing artifacts.

## Testing

- `uv run --extra dev pytest tests/unit/test_hub_artifacts.py`
- `uv run ruff check .`

### DIFF
--- a/agent/tools/sandbox_tool.py
+++ b/agent/tools/sandbox_tool.py
@@ -374,18 +374,6 @@ async def _create_sandbox_locked(
         create_latency_s=int(_t.monotonic() - _t_start),
     )
 
-    # Set a descriptive title (template title is inherited on duplicate)
-    from huggingface_hub import metadata_update
-
-    await asyncio.to_thread(
-        metadata_update,
-        sb.space_id,
-        {"title": "ml-intern sandbox"},
-        repo_type="space",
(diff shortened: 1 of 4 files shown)


agent/tools/sandbox_tool.py
```diff
@@ -374,18 +374,6 @@ async def _watch_cancel():
         create_latency_s=int(_t.monotonic() - _t_start),
     )
 
-    # Set a descriptive title (template title is inherited on duplicate)
-    from huggingface_hub import metadata_update
-
-    await asyncio.to_thread(
-        metadata_update,
```
lewtun, line 381: We skip this because it forces a restart just to change a README

PR #243, opened 2026-05-07
fix: catch tool exceptions to prevent corrupted agent context

Olexandr88:
Unexpected exceptions in _exec_tool could leave the agent context in an invalid state an assistant message with tool_calls committed but no corresponding tool results. Wrapped call_tool in try/except to always return a valid result tuple.


lewtun:

PR #245, opened 2026-05-08
Create Hub artifact collections lazily

lewtun:
## Summary

This changes Hub artifact collection creation from eager session startup to lazy creation during artifact registration. Sessions that never produce Hub artifacts will no longer leave behind empty private collections.

## What changed

- Removed the eager collection startup path from frontend session creation, session restore, and existing-session access.
- Removed eager collection scheduling from the core agent loop.
- Kept collection creation inside artifact registration, immediately before adding the first collection item.
- Made collection registration report success only after `add_collection_item` succeeds, so failed collection adds are retried instead of cached as registered.
- Updated tests that previously asserted eager collection creation to assert no collection task/slug is attached on session create or restore.

## Verification

- `uv run ruff check .`
- `uv run pytest tests/unit/test_hub_artifacts.py tests/unit/test_session_manager_persistence.py`

## Notes

Existing empty collections are intentionally out of scope; this prevents new empty collections from being created.


lewtun:
Addressed the P2 about child-process `sitecustomize` losing the warmed collection slug.

Update pushed in `107ee0f`:
- generated `sitecustomize` now caches the lazily created collection slug in a stable per-session temp-file path, with `ML_INTERN_ARTIFACT_COLLECTION_CACHE` as an override for tests/custom environments;
- fresh child bootstraps read that cache before calling `create_collection`, so later child Python processes in the same job/sandbox/local environment reuse the slug while preserving lazy creation;
- added a regression test that simulates two fresh bootstraps and verifies `create_collection` is called once while both artifacts are added.

Verification:
- `uv run ruff check .`
- `uv run pytest tests/unit/test_hub_artifacts.py tests/unit/test_session_manager_persistence.py`

PR #246, opened 2026-05-08
Fix GPU sandbox hardware OAuth failure

lewtun:
## Summary

Fixes GPU sandbox creation in the deployed Space by avoiding the redundant post-duplicate `/hardware` request that can fail with a 401 for fresh private Spaces created with HF OAuth tokens.

## Root Cause

`Sandbox.create()` already passes the requested hardware and sleep time to `HfApi.duplicate_space()`. After duplication, we were making a second `request_space_hardware()` call against `/api/spaces/{repo}/hardware`. In the deployed OAuth flow, that endpoint can return `401 Repository Not Found / Invalid username or password` for a newly-created private Space even though the duplicate request succeeded.

This made GPU sandbox creation fail after the Space was already created, and the model surfaced the error as an authentication problem.

## Changes

- Removed the post-duplicate hardware request and retry path.
- Rely on `duplicate_space(..., hardware=..., sleep_time=...)` as the single source of truth for sandbox hardware provisioning.
- Updated sandbox tests to assert GPU hardware and sleep time are configured during duplication and that no standalone hardware request is made.

## Verification

- `uv run pytest tests/unit/test_sandbox_private_spaces.py tests/unit/test_sandbox_auto_start.py tests/unit/test_sandbox_api_auth.py`
- `uv run ruff check --exclude train_smollm2.py .`
- `uv run ruff format --check --exclude train_smollm2.py .`

The `train_smollm2.py` exclusion is only for an unrelated untracked local file.

### DIFF
--- a/agent/tools/sandbox_client.py
+++ b/agent/tools/sandbox_client.py
@@ -65,7 +65,6 @@ MAX_TIMEOUT = 1200
 WAIT_TIMEOUT = 600
 WAIT_INTERVAL = 5
 API_WAIT_TIMEOUT = 180
-HARDWARE_REQUEST_TIMEOUT = 60
 CPU_BASIC_HARDWARE = "cpu-basic"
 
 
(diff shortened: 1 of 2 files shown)


lewtun:
Addressed the Claude review in `134fee8`:

- clarified the sandbox client comment to say we rely on `duplicate_space` to honor `sleepTimeSeconds` for upgraded hardware
- made `cpu-basic` sleep-time behavior explicit in the tool log because Hub auto-sleep is fixed on that tier
- added coverage that no sleep-time log is emitted when `sleep_time=None`
- added coverage for the `cpu-basic + sleep_time` case, including no fallback `/hardware` call

Verification:

- `uv run pytest tests/unit/test_sandbox_private_spaces.py tests/unit/test_sandbox_auto_start.py tests/unit/test_sandbox_api_auth.py`
- `uv run ruff check --exclude train_smollm2.py .`
- `uv run ruff format --check --exclude train_smollm2.py .`

PR #248, opened 2026-05-08
Align CLI help output

lewtun:
## Summary

- Replace the hard-coded `/help` text block with row-based formatting that pads command names and arguments consistently.
- Keep current commands, including `/resume`, aligned in one stable description column.
- Shorten the longest help descriptions so the default terminal width does not wrap the table.

## Testing

- `uv run pytest tests/unit/test_cli_rendering.py -q`
- `uv run ruff check agent/utils/terminal_display.py tests/unit/test_cli_rendering.py`
- `uv run ruff format --check agent/utils/terminal_display.py tests/unit/test_cli_rendering.py`

### DIFF
--- a/agent/utils/terminal_display.py
+++ b/agent/utils/terminal_display.py
@@ -468,30 +468,51 @@ HELP_ROWS: tuple[tuple[str, str, str], ...] = (
     ("/quit", "", "Exit"),
 )
 
-_HELP_COMMAND_WIDTH = max(len(command) for command, _, _ in HELP_ROWS)
-_HELP_ARGS_WIDTH = max(len(args) for _, args, _ in HELP_ROWS)
+
+def _help_column_widths(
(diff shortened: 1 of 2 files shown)


lewtun:
Addressed both review notes in `353cfca`:

- The help rendering test now uses `terminal_display._THEME`, matching production console styling.
- Help text is now formatted at print time via `format_help_text()`, so widths are recomputed from the current rows instead of freezing `HELP_TEXT` at import time.
- Added a regression test with a synthetic longer row to prove future wider rows still align.

Checks:

- `uv run pytest tests/unit/test_cli_rendering.py -q`
- `uv run ruff check agent/utils/terminal_display.py tests/unit/test_cli_rendering.py`
- `uv run ruff format --check agent/utils/terminal_display.py tests/unit/test_cli_rendering.py`

PR #260, opened 2026-05-15
feat(research): inject goal anchor every N iterations to prevent drift

Meur3ault:
# Re-anchor research agent to original task every N iterations

## Problem

The research sub-agent runs up to 60 iterations. After a long chain of tool
calls the original task ends up at position 1 in the message list, buried under
dozens of assistant/tool message pairs. In practice the model starts
summarising whatever it found most recently rather than what it was asked to
find.

This is specific to the research sub-agent: the task is a single fixed string
set once at the start, there are no follow-up user messages to re-anchor it,
and the loop runs fully autonomously. The main agent doesn't have this problem
because each user turn naturally resets focus.

## Fix

Every 10 iterations, append a `[SYSTEM: GOAL ANCHOR]` user message containing:

- the original task verbatim
- a snippet of any thinking text the model produced alongside the previous tool
  calls (capped at 500 chars, used as a cheap "progress so far" without an
  extra LLM call)

The anchor fires after doom-loop detection and before the context-budget check,
so all three nag mechanisms stay independent.

## Testing

Unit tests cover the injection boundaries, periodic repeat at 2N, doom-loop
coexistence, and a structural test that verifies the anchor is the most recent
user message before the LLM call with exactly N tool-result messages between it
and the original task statement.

For the actual capability claim I ran a live A/B test (`tests/integration/
test_research_anchor_eval.py`) against Llama-3.1-8B via HF router: injected 15
off-topic time-series tool results to simulate drift, then compared summaries
with and without the anchor.

```
A (no anchor):  score = -3   off-task keywords dominated
B (with anchor): score = +4   model re-engaged with LoRA fine-tuning task
```

The eval asserts both that drift occurs without the anchor and that it's
corrected with it. Enable with `ML_INTERN_LIVE_LLM_TESTS=1 HF_TOKEN=...`.

### DIFF
--- a/agent/tools/research_tool.py
+++ b/agent/tools/research_tool.py
@@ -28,6 +28,11 @@ logger = logging.getLogger(__name__)
 _RESEARCH_CONTEXT_WARN = 170_000  # 85% of 200k
 _RESEARCH_CONTEXT_MAX = 190_000
 
+# N-turn goal anchor: re-inject the original task every N iterations so the
+# agent stays anchored even after long tool-call chains.
+_RESEARCH_FACT_INTERVAL = 10
+_RESEARCH_FACT_SUMMARY_MAX = 500
+
 # Tools the research agent can use (read-only subset)
 RESEARCH_TOOL_NAMES = {
     "read",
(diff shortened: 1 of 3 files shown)


lewtun:

PR #263, opened 2026-05-16
Security: Default API key in local model configuration

tuanaiseo:
## Problem

In agent/core/local_models.py, the default API key is set to 'sk-local-no-key-required' which provides a non-empty fallback rather than requiring explicit configuration. This could lead to unintended connections to local LLM servers without proper authentication.

**Severity**: `medium`
**File**: `agent/core/local_models.py`

## Solution

Consider using None as default and requiring explicit configuration, or at minimum document this behavior clearly and ensure it's only used in development contexts.

## Changes

- `agent/core/local_models.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced


lewtun:
Closing this as `close-not-a-bug`.

The patch removes a documented optional local-auth fallback that tests currently assert, without showing a concrete vulnerability beyond localhost OpenAI-compatible convenience behavior.

A new issue with a concrete failing scenario would be useful if there is still a real bug here.

PR #266, opened 2026-05-17
Fix :Remove broken HF MCP server from default CLI config

Oscar963852:
The hf-mcp-server entry causes the agent to hang on startup. Removing it so the CLI works out of the box.


lewtun:

PR #267, opened 2026-05-17
Gate sandbox-only prompt sections on local_mode

Abhilekh-Meda:
## Summary

The system prompt asserted sandbox-only details (sandbox_create,/app/ paths, "session sandbox") unconditionally, then a postscript in ContextManager appended a "actually ignore that" paragraph in local mode. The agent had to hold two contradictory frames and pick the right one at runtime.

This PR moves the mode split into the Jinja template:
- Five sandbox-only sections in `system_prompt_v3.yaml` are wrapped in `{% if not local_mode %}` / `{% else %}` gates, each with a paralllel local-mode variant carrying the equivalent information.
- `ContextManager._load_system_prompt` passes `local_mode` and `cwd` into `template.render()`.
- The 16 line postscript is deleted

## Verification

- **Sandbox rendering is identical to the pre-change output**. Non-local users see zero change.
-  Local rendering carries the equivalent info the postscript used to (no sandbox, working directory, no `sandbox_create`, no `/app/` paths). Now natively in the prompt instead of as a contradicting postscript.
- All unit tests pass.

##Out of scope

There is similar pattern with autonomous / headless mode asking the LLM to self assess. The autonomous gate would need new plunmbing (`autonomous: bool`).

## Test plan
- [x] Sandbox render unchanged (diff checked)
- [x] Local render contains no `sandbox_create` / `/app/` references
- [x] Full unit test suite passes
- [x] Manual smoke test: confirm normal startup

### DIFF
--- a/agent/context_manager/manager.py
+++ b/agent/context_manager/manager.py
@@ -255,28 +255,20 @@ class ContextManager:
         # Get HF user info from OAuth token
         hf_user_info = _get_hf_username(hf_token)
 
-        template = Template(template_str)
-        static_prompt = template.render(
-            tools=tool_specs,
-            num_tools=len(tool_specs),
-        )
-
-        # CLI-specific context for local mode
         if local_mode:
             import os
 
             cwd = os.getcwd()
-            local_context = (
-                f"\n\n# CLI / Local mode\n\n"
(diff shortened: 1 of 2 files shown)


lewtun:

PR #269, opened 2026-05-22
fix(tools): harden bash tool argument handling

SuperMarioYL:
## What

`_bash_handler` (the bash tool) had two argument-handling defects. This PR hardens both and adds regression tests.

### 1. A string `timeout` crashed the agent turn

`timeout` is computed *before* the handler's `try` block:

```python
timeout = min(args.get("timeout") or DEFAULT_TIMEOUT, MAX_TIMEOUT)
```

The bash tool schema declares `timeout` as an integer, but models occasionally emit schema-typed integers as JSON strings (`"120"`). `min("120", 36000)` then raises an uncaught `TypeError` that aborts the whole turn instead of returning a recoverable tool error:

```
TypeError: '<' not supported between instances of 'int' and 'str'
```

The value is now coerced with `int()`, falling back to `DEFAULT_TIMEOUT` on a non-numeric input. A non-positive value (e.g. `"-5"` or `"0"`) — which would otherwise make `subprocess.run` time out instantly with a misleading message — also falls back to the default.

### 2. An invalid `work_dir` surfaced a raw `OSError`

A non-existent `work_dir` reached `subprocess.run(cwd=...)` and surfaced a raw `FileNotFoundError` string. It is now validated up front with a clear, actionable message so the model can correct the path.

## Tests

Adds `tests/unit/test_local_tools.py` (10 tests) covering string / missing / integer / oversized / oversized-string / non-numeric / non-positive `timeout`, and missing / file / valid `work_dir`. The string-`timeout` test is red on `main` and green here.

`uv run pytest tests/unit/test_local_tools.py` -> 10 passed. `ruff check` / `ruff format --check` clean.

## Notes

Behavior for valid integer input is unchanged. This is the bash-tool sibling of the read-tool coercion in #110 — it touches a different function (`_bash_handler`, not `_read_handler`) and does not overlap that PR.

### DIFF
--- a/agent/tools/local_tools.py
+++ b/agent/tools/local_tools.py
@@ -112,9 +112,13 @@ async def _bash_handler(
         return f"work_dir is not an existing directory: {work_dir}", False
     # Providers occasionally emit the schema-typed integer `timeout` as a
     # string; coerce it so a string value can't raise an uncaught TypeError.
+    # A non-positive value would make subprocess.run time out instantly, so
+    # fall back to the default in that case too.
     try:
         timeout = int(args.get("timeout") or DEFAULT_TIMEOUT)
     except (TypeError, ValueError):
+        timeout = DEFAULT_TIMEOUT
+    if timeout <= 0:
         timeout = DEFAULT_TIMEOUT
     timeout = min(timeout, MAX_TIMEOUT)
     try:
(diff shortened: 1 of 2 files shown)


lewtun:

PR #284, opened 2026-06-03
Set Sonnet as default premium model

lewtun:


lewtun:
Addressed the review feedback in 97d6d42: premium plan gating now runs before the already-counted session shortcut, downgraded users get a regression test, restored pro-only sessions are blocked in the frontend send path, and the HF Router ID punctuation is documented. Checks: pytest tests/unit/test_agent_model_gating.py tests/unit/test_llm_params.py tests/unit/test_cli_local_models.py, npm run --prefix frontend build, uv run ruff check ., uv run ruff format --check .

PR #295, opened 2026-06-07
keep per-token events out of saved trajectory

RyanRana:
I was using ML Intern, and then on my own observability tools, I noticed that when there were lots of turns and streamed outputs, the pauses per turn were getting longer and longer. I checked, and it was the trajectory save that was increasing and not the snapshot. Then I built a small test and noticed that it was the assistant chunk in log events that streams tokens and gets fully reserialized after, so all I did stop persisting per-token assistant_chunk events (they're still streamed live, and the real text lives in context_manager.items)


RyanRana:
agent/core/session.py is the fix, the new file is just the test i used.

lewtun:

PR #308, opened 2026-06-08
Recover chat sends after fetch failures

lewtun:
## Summary
- recover chat POST fetch failures by probing session state and reconnecting to live SSE when backend accepted the turn
- hydrate backend messages after completed turns so transient browser/network failures do not surface as hard message failures
- keep aborts and HTTP/provider errors on their existing paths

## Checks
- npm run build
- npm run lint (passes with existing warnings in main.tsx and logger.ts)
- uv run ruff check .
- uv run ruff format --check .


lewtun:
Addressed the review in 6510d69:

- P0: recovery hydration now only overwrites chat state when backend UI history has at least the current message count, preserving the optimistic user message when the POST was not accepted.
- P1: recoverable fetch classification is narrowed to known browser/network failure signatures instead of every TypeError or broad fetch/connection substrings.

Checks:
- npm run build
- npm run lint (existing warnings only)
- uv run ruff check .
- uv run ruff format --check .

lewtun:
Addressed the redundant-fetch P2 in 05d4423:

- The transport now passes the already-fetched session info into `onRecoverMessages` when the recovery probe shows the backend is idle.
- `useAgentChat` uses that supplied session info and only fetches `/messages` in that idle recovery path, avoiding the duplicate `/api/session/{id}` request.
- If the probe showed `is_processing: true` but live SSE reconnect fails, the callback still refetches `/session` because the turn may have completed between the probe and hydration.

Checks repeated:
- npm run build
- npm run lint (existing warnings only)
- uv run ruff check .
- uv run ruff format --check .

Browser sanity check: local app reloads at http://127.0.0.1:5173/ with title `ML Intern` and populated root content.

PR #313, opened 2026-06-10
Enforce YOLO cap across session usage

lewtun:
## Summary
- Adds a session-scoped YOLO budget ledger for LLM calls, HF jobs, and sandboxes.
- Fails closed for unknown-cost LLM/tool spend and re-checks manual approvals while YOLO is enabled.
- Adds sandbox budget renewal/teardown and updates frontend blocked-state copy.

## Tests
- uv run --extra dev --frozen ruff check .
- uv run --extra dev --frozen ruff format --check .
- uv run --extra dev --frozen pytest tests/unit -q
- npm run build


lewtun:
Addressed the review items in f8a8fb1.

- P0: `reconcile_budget_reservation` now treats zero actual cost as unknown by default, so unpriced HF Router `completion_cost == 0.0` no longer clears the reserved LLM estimate. Jobs and sandboxes opt in to measured zero-cost reconciliation with `allow_zero_actual=True`.
- P1 output bound: YOLO injected LLM output bound now defaults to 32768 and is capped by remaining router context when available. Main-loop responses that hit an injected YOLO bound now emit a system tool log so truncation is visible.
- P1 catalog failure: router catalog fetch failures now use a short failure retry TTL, keep stale cache usable, expose `last_fetch_error`, and produce a distinct fail-closed reason from genuinely missing model prices.
- P2: removed the unreachable YOLO-budget `complete_turn` branch and unused pending fields.

Validation:
- `uv run --extra dev --frozen ruff check .`
- `uv run --extra dev --frozen ruff format --check .`
- `uv run --extra dev --frozen pytest tests/unit -q` (478 passed)
- `npm run build`

lewtun:
Updated per latest request: inference-provider/LLM calls no longer pre-estimate or reserve YOLO budget. They run first, then the session checks fresh current-session usage and raises the synthetic yolo_budget warning only after the cap is reached. Jobs and sandboxes still keep the existing upfront bounded reservations.

Also removed the dead LLM pre-estimation helper path, preserved completed text responses via complete_turn yolo warnings, and fixed the UI so unknown next estimates render as Unknown instead of $0.00.

Verified:
- uv run ruff check .
- uv run ruff format --check .
- uv run pytest tests/unit -q
- npm run build

PR #310, opened 2026-06-09
Add session usage threshold approvals

lewtun:
## Summary
- add synthetic usage-threshold approvals at $5, $10, $20, $40, and beyond
- pause the agent at safe loop boundaries and reuse the existing approval SSE flow
- persist and hydrate pending usage warnings across session restore and refresh
- render usage warnings with Continue and Stop here actions in the existing approval card UI

## Checks
- uv run --frozen --extra dev pytest tests/unit/test_usage.py tests/unit/test_session_manager_persistence.py tests/unit/test_auto_approval_policy.py tests/unit/test_usage_threshold_approvals.py
- uv run --frozen --extra dev ruff check .
- uv run --frozen --extra dev ruff format --check .
- cd frontend && npm run build

PR #320, opened 2026-06-12
Evict pinned sessions and retry Mongo init in the idle reaper

lewtun:
## Why

Production incident on the deployed Space: users hit **"maximum of 10 live sessions"** hours after their last activity. Diagnosis from the Space logs and code:

1. The idle reaper **permanently skips** sessions with `pending_approval` set. Since #310/#313, usage-threshold and YOLO-cap prompts are auto-created **at turn end with no user action**, so an unattended user accumulates unreapable sessions until they fill all 10 slots. Unanswered tool-permission prompts pin slots the same way.
2. A turn stuck `is_processing` pins its slot forever — there was no hung-turn detection.
3. `MongoSessionStore.init()` gives up permanently after one failed ping at boot; the reaper no-ops while the store is disabled, so a single Mongo blip at startup silently killed reaping until the next restart.
4. The reaper logged nothing about skipped sessions, so none of this was diagnosable from logs.

## What changed

**Per-session verdict classifier** (`_reaper_verdict`) replaces the single idle predicate; `_reap_one` re-checks the same verdict under the lock, so a session whose state changes mid-reap (e.g. prompt answered) aborts cleanly:

| State | Eviction window | Clock |
|---|---|---|
| Plain idle | 15 min (`REAPER_IDLE_MINUTES`, unchanged) | `last_active_at` |
| Usage-threshold / YOLO ack prompt | 15 min | `last_active_at` |
| Real tool-permission prompt | 60 min (`REAPER_TOOL_APPROVAL_IDLE_MINUTES`) | `last_active_at` |
| Processing, event-silent | 15 min (`REAPER_STALLED_MINUTES`) | new `last_event_at` |
| Processing, emitting events | never (healthy long run) | — |

- **`last_event_at`** is stamped as events flow through the `EventBroadcaster` (every chunk, tool update, heartbeat), so legitimate long-running jobs are spared while hung turns go silent and get evicted. Stalled evictions set the cooperative interrupt flag before `task.cancel()` and ignore queued submissions (a hung turn never drains its queue; queue depth is logged).
- Evicted sessions stay **fully resumable**: snapshots persist `status="active"` with `runtime_state="waiting_approval"`/`"idle"` (never `"processing"`), and pending approvals round-trip through the existing serialize/restore path so prompts remain answerable after restore.
- **Mongo retry**: the sweep calls `store.maybe_reconnect()` while the store is disabled, re-attempting `init()` every 5-min sweep instead of staying dead until restart.
- **Observability**: one per-sweep counter line (`Reaper sweep: evicted_idle=… evicted_pending_ack=… evicted_stalled=… skipped_processing=… aborted=…`, only when nonzero), a warning per stalled eviction, and the per-user 503 now appends `Currently held: N awaiting your approval, M still processing, K idle.`

## Accepted tradeoffs

- A single silent tool call blocking >15 min with no events is treated as hung and evicted (session stays resumable).
- Evicting a session parked on a real tool approval destroys its sandbox; on restore the sandbox is recreated empty (pre-existing behavior for restored pending-approval sessions, now more common at the 60-min mark).

## Testing

- `tests/unit/test_session_reaper.py`: 32 tests (13 new/reworked) covering ack/tool-approval eviction windows, stalled-turn eviction + healthy-long-turn sparing, `last_event_at` stamping, verdict-mismatch aborts, Mongo retry recovery/no-op, capacity-message breakdown, and sweep-log gating.
- Full unit suite: **510 passed**. `ruff check` + `ruff format --check` clean.

### DIFF
--- a/agent/core/session_persistence.py
+++ b/agent/core/session_persistence.py
@@ -265,6 +265,34 @@ class MongoSessionStore(NoopSessionStore):
                 raise RuntimeError("session store not ready")
             return
         now = _now()
+        if raise_on_error:
+            await self._write_snapshot_messages(
+                session_id=session_id,
+                messages=messages,
+                now=now,
+                raise_on_error=True,
+            )
+            await self.upsert_session(
+                session_id=session_id,
+                user_id=user_id,
+                model=model,
+                title=title,
+                created_at=created_at,
+                runtime_state=runtime_state,
+                status=status,
+                message_count=len(messages),
+                turn_count=turn_count,
+                pending_approval=pending_approval,
+                notification_destinations=notification_destinations,
+                usage_window_started_at=usage_window_started_at,
+                inference_billing_session_id=inference_billing_session_id,
+                auto_approval_enabled=auto_approval_enabled,
+                auto_approval_cost_cap_usd=auto_approval_cost_cap_usd,
+                auto_approval_estimated_spend_usd=auto_approval_estimated_spend_usd,
+                usage_warning_next_threshold_usd=usage_warning_next_threshold_usd,
+            )
+            return
+
         await self.upsert_session(
             session_id=session_id,
             user_id=user_id,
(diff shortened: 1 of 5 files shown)


lewtun:

PR #324, opened 2026-06-14
Bill inference to an org + fix per-project .env loading

FabienDanieau:
Two independent fixes, prompted by hitting 402 — monthly Inference Providers credits exhausted on personal credits with no way to redirect billing.

feat: HF_BILL_TO env var — All HF Router calls (main agent, research sub-agent, compaction) billed the token owner's personal allowance. Set HF_BILL_TO=<org> to attach an X-HF-Bill-To header and charge that org's credits instead. Mirrors huggingface_hub's bill_to=; local OpenAI-compatible endpoints never get it.

fix: launch-directory .env ignored — load_config meant to read a .env from the directory you run ml-intern from, but bare load_dotenv() makes find_dotenv walk up from config.py's own location, never the launch CWD — so a per-project .env was silently dropped (only the repo's own .env loaded). Now resolved explicitly with find_dotenv(usecwd=True). Precedence unchanged: repo .env still wins on conflicts.

Testing

- New test_llm_params.py cases: header set/omitted by env, whitespace trimmed, never applied to local models.
- Verified the X-HF-Bill-To header reaches the real request to router.huggingface.co.
- Verified a launch-directory .env now populates HF_BILL_TO.
- ruff clean; test_llm_params.py (25) and test_config.py (7) pass.


lewtun:

PR #326, opened 2026-06-14
Add session titles, /rename, and a type-to-filter resume picker

### LINKED ISSUE #325: [Feature] Add session naming with /rename and improve resume picker UX
## Summary
  I'd like to make local CLI sessions easy to find and resume. Today each session
  is identified only by a random UUID, stored in a directory relative to wherever
  the CLI was launched, and `/resume` lists sessions in a confusing order with weak
  previews. There is no way to name a conversation.

  ## Motivation
  ChatGPT and Claude both auto-title conversations and let you rename them. That
  single feature is what makes past sessions findable. In ml-intern there is no
  session title anywhere in the data model (`session_id = str(uuid.uuid4())`,
  `agent/core/session.py:141`) and no `/rename` command (see `HELP_ROWS` in
  `agent/utils/terminal_display.py:455`). When you run `/resume` you get rows like
  `2026-05-03 14:41  2898bada  2 msgs  ...` with previews such as `openai/gpt-5.5`
  or `(no user prompt preview)`, and the list does not even appear sorted by date.

  This follows the same spirit as #256 ("Add `/clear` and `/new` commands").

  ## Proposed Design

  ### 1. Conversation naming and `/rename` (primary feature)
  - Add a `session_title` field to the session model, persisted in the log JSON.
  - Auto-title from the conversation after the first turn, ChatGPT/Claude style.
  - Add a `/rename <name>` command to set or override the title at any time.
  - Show the title in `/resume` and in the saved filename.

  ### 2. Standardize the session storage location
  `DEFAULT_SESSION_LOG_DIR = Path("session_logs")` (`agent/core/session.py:30`) is
  relative to the launch directory, so sessions scatter and `/resume` only sees the
  current directory's folder.
  - Store under XDG: `$XDG_DATA_HOME/ml-intern/sessions`
    (default `~/.local/share/ml-intern/sessions`), with a config/env override.
  - Still read a legacy local `session_logs/` if present, for backward compat.

  ### 3. Readable filenames
  Files are named `session_<uuid4>_<YYYYMMDD_HHMMSS>.json`
  (`agent/core/session.py:614`), which are impossible to scan by eye.
  - Rename to a readable, chronologically-sortable scheme that includes the title,
    for example `20260614_164207_finetune-llama-on-squad.json`.

 ### 4. Make `/resume` an interactive picker and fix order and previews
  Today `/resume` prints a numbered list and asks you to type a number
  ("Select session number"). Claude Code style arrow-key selection is far more
  convenient.
  - Replace the type-a-number flow with an interactive picker: arrow keys to move,
    Enter to select, Esc to cancel. This needs no new dependency since the CLI
    already uses `prompt_toolkit` (`agent/main.py:23`); a `radiolist_dialog` or a
    custom key-bound list works. Keep `/resume <index|id|path>` as a non-interactive
    fallback.
  - The list is sorted by file `mtime` (`agent/core/session_resume.py:106`) but each
    row displays `session_end_time` (`session_resume.py:111`). Heartbeat re-saves
    bump `mtime` but not the end-time, so the order looks scrambled. Sort and
    display should use the same timestamp.
  - `_first_user_preview` (`session_resume.py:60`) returns the first `user`-role
    message, which is often a `/model` swap or empty. Once titles exist, show the
    title; otherwise prefer the first real user prose.

  ## Scope
  CLI only (`agent/`). Web/backend MongoDB persistence is out of scope.

mohitpaddhariya:
## Summary

Implements #325. Makes local CLI sessions easy to name, find, and resume.

- **Session titles + `/rename` + auto-title** — every session gets a short human-readable title (auto-generated from the conversation after the first turn, with a pure-Python fallback), and `/rename <name>` lets you set it explicitly.
- **Standardized storage** — session logs move from a CWD-relative `./session_logs` to an XDG path (`$XDG_DATA_HOME/ml-intern/sessions`), overridable via config or `ML_INTERN_SESSION_DIR`, with a legacy fallback so existing checkouts are undisturbed.
- **Readable filenames** — `session_<slug>_<uuid8>_<timestamp>.json` instead of a bare UUID.
- **Type-to-filter resume picker** — `/resume` opens an arrow-key picker (filter by title/preview/model), shows titles, fixes the sort order, and dedupes a resumed conversation so it never appears twice. `/resume <index|id|path>` still works as a fast path.

### 1. Titling: `/rename` + auto-title

```mermaid
flowchart TD
  A["First user turn completes"] --> B{"Already titled<br/>or renamed?"}
  B -- no --> C["Auto-title: ask the model<br/>for a 3-6 word title"]
  C -- success --> T["session_title"]
  C -- "error / empty" --> D["Fallback: slug of<br/>first message"]
  D --> T
  B -- yes --> T
  RN["/rename name"] --> T
  T --> P["persist_title()"]
  P --> P1["rename log in place +<br/>refresh title inside the JSON"]
  P --> P2["or save a fresh titled log<br/>if none exists yet"]
```

### 2. Storage: `resolve_session_log_dir(config)`

```mermaid
flowchart TD
  S0["resolve_session_log_dir(config)"] --> S2{"config.session_log_dir set?"}
  S2 -- yes --> U1["use it"]
  S2 -- no --> S3{"$ML_INTERN_SESSION_DIR set?"}
  S3 -- yes --> U2["use it"]
  S3 -- no --> S4{"./session_logs exists in cwd?"}
  S4 -- yes --> U3["use legacy ./session_logs"]
  S4 -- no --> U4["XDG: ~/.local/share/ml-intern/sessions"]
```

### 3. Filenames

```mermaid
flowchart TD
  N0["save a session log"] --> N1{"has a usable title?"}
  N1 -- yes --> N2["session_&lt;slug&gt;_&lt;uuid8&gt;_&lt;timestamp&gt;.json"]
  N1 -- no --> N3["session_&lt;uuid&gt;_&lt;timestamp&gt;.json (legacy)"]
```

### 4. Resume: `/resume`

```mermaid
flowchart TD
  R1["/resume"] --> R2["list_session_logs from resolved dir"]
  R2 --> R3["sort by one tz-aware timestamp"]
  R3 --> R4["dedupe by session_id (keep newest)"]
  R4 --> R5["type-to-filter arrow-key picker"]
  R5 --> R6["restore_session_from_log"]
  R6 --> R7["carry session_title across resume"]
```

## Key design points

- **One canonical timestamp** for both sorting and display (`session_end_time` → `session_start_time` → file mtime), tz-aware, so the list is genuinely newest-first and labels never look scrambled.
- **Dedupe by `session_id`** — a resumed-and-continued conversation forks the save path into a new-timestamp file; the listing collapses these to the newest entry so the same conversation never shows twice.
- **`persist_title()`** — setting a title (auto or `/rename`) renames the active log in place *and* refreshes the title stored inside the JSON, or saves a fresh titled log when none exists yet (e.g. right after a resume), so `/resume` reflects the new name immediately.
- **One resolver** (`resolve_session_log_dir`) is used by every reader and writer, so reads and writes can never diverge.
- Secret-like tokens are stripped from titles before they reach a filename or the JSON.

## Testing

- New unit tests: `test_title.py`, `test_session_title.py`, `test_session_dir.py`, `test_session_filename.py`, `test_session_picker.py`, plus additions to `test_session_resume.py`.
- Covers auto-title + fallback, `/rename`, XDG resolution + precedence, titled/legacy filenames, in-place rename + title refresh, sort/display agreement, and dedupe (including not merging genuinely distinct sessions).
- Full unit suite passes locally.

## Screenshots

**Auto-generated title after the first turn**

Auto-title message shown after the first turn

**`/resume` — type-to-filter picker**

New /resume type-to-filter picker

**`/rename` — renaming a session**

/rename command output

Closes #325


mohitpaddhariya:
Hi @lewtun and @akseljoonas, this PR (#326) implements #325 and includes:

* Session titles
* `/rename`
* Automatic title generation
* XDG-compliant session storage
* Readable session filenames
* A type-to-filter `/resume` picker

The full write-up, diagrams, and screenshots are in the PR description, and the changes are covered by unit tests.

Whenever you have a chance, I'd appreciate a review and merge consideration. Happy to make any adjustments if needed.

**Note:** The red **"Claude PR Review"** check is expected to fail on this PR because it originates from a fork and only has read-only token permissions. It does **not** indicate a problem with the code or tests.

mohitpaddhariya (commented):
.

mohitpaddhariya:
Pushed a follow-up commit (`fd3ae7c`) that hardens this PR based on a self-review pass before merge. Everything stays within the existing scope — titles / `/rename` / auto-title, storage, and resume — no new surface.

**Titling**
- Fixed a race in the fire-and-forget auto-title task: it now snapshots a per-conversation epoch and bails if `/new` or `/resume` rotated the conversation during the title LLM call, so a stale title can't be written onto a different session.
- Auto-title now also fires from the usage-threshold / YOLO / approval-abandon completion paths, so a first turn that paused for an approval still gets titled instead of staying permanently untitled.
- The title task keeps a strong reference now (same pattern as `_heartbeat_tasks` in `telemetry.py`), so it can't be GC'd mid-await.

**Secret stripping** (the "secret-like tokens are stripped from titles" design point)
- Titles now route through the shared `redact.scrub_string` on the auto-title path, the `/rename` path, and the persisted JSON/filename. The old 30-char heuristic missed structured secrets the project redactor catches (AWS key ids, `Bearer` tokens, `NAME=value` dumps); those no longer reach the filename slug or the `session_title` field. Scope is local-disk only — `session_title` isn't in either upload payload.
- Titles are `escape()`d before being rendered as Rich markup.

**Storage** (the "one resolver" design point)
- `/resume` now union-reads a legacy `./session_logs` alongside the resolved dir, so sessions created before the XDG move stay visible regardless of launch cwd (non-destructive — writes still go to the resolved dir).
- Dropped the divergent cwd-relative directory defaults that bypassed `resolve_session_log_dir`, keeping the single-resolver invariant honest.

**Concurrency**
- Guarded the title rename + trajectory save with a lock so a heartbeat save on the worker thread can't resurrect a pre-title log file.

Full unit suite passes locally (542).

lewtun:

PR #349, opened 2026-07-11
[codex] Replace curl subprocess with huggingface_hub in _get_hf_username

rahulrao85:
## Summary

Replace the `subprocess.run(['curl', ...])` call in `_get_hf_username()` with `huggingface_hub.HfApi().whoami(token=hf_token)`.

## Motivation

- `curl` is not available on Windows by default — breaks the agent on Windows
- The project already depends on `huggingface_hub` — no new dependency needed
- `HfApi().whoami()` handles IPv6 Happy Eyeballs internally (the reason `curl -4` was used)

## Validation

- `from huggingface_hub import HfApi` confirmed importable with the project's `huggingface-hub>=1.12.0` dependency
- Function signature unchanged: `_get_hf_username(hf_token: str | None = None) -> str`
- Maintains same fallback behavior: returns `"unknown"` on any failure

### DIFF
--- a/agent/context_manager/manager.py
+++ b/agent/context_manager/manager.py
@@ -21,56 +21,22 @@ from agent.core.prompt_caching import (
 
 logger = logging.getLogger(__name__)
 
-_HF_WHOAMI_URL = "https://huggingface.co/api/whoami-v2"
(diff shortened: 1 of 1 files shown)


lewtun:

PR #350, opened 2026-07-11
[codex] Use atomic write in _update_upload_status to prevent file corruption

rahulrao85:
## Summary

Replace the `seek(0) → json.dump → truncate()` in-place rewrite pattern in `_update_upload_status()` with a temp-file + `os.replace()` atomic write pattern.

## Problem

The previous implementation used:
```python
f.seek(0)
json.dump(data, f, indent=2)
f.truncate()
```

If `json.dump()` raised an exception mid-write (e.g. serialization error for a non-serializable value), the file was left with partially-overwritten data at a wrong truncation position — corrupting the session JSON file permanently. Both the org and personal uploaders use this same function, so a crash in either process would corrupt the shared session file.

## Fix

1. Read the session file under the exclusive lock (unchanged)
2. Write the full JSON to a `.tmp` sibling file using `json.dump`
3. Atomically swap with `os.replace(tmp_path, session_file)`

If `json.dump` crashes, the `.tmp` file is cleaned up and the original `session_file` remains untouched. `os.replace()` is atomic on POSIX and modern Windows (NTFS atomic rename), so readers always see a consistent file.

## Validation

- Same locking semantics preserved (exclusive lock protects the read)
- `os.replace` is used instead of `shutil.move` for cross-filesystem safety (ensures atomic rename when source and dest are on the same filesystem)
- Temp file cleanup in `except` block prevents orphaned `.tmp` files

### DIFF
--- a/agent/core/session_uploader.py
+++ b/agent/core/session_uploader.py
@@ -349,10 +349,14 @@ def _update_upload_status(
     The org and personal uploaders run as separate processes against the same
     local session JSON file. Re-read under an exclusive lock so one uploader
     cannot clobber fields written by the other.
+
+    Uses a temp-file + os.replace pattern to prevent file corruption if
+    json.dump crashes mid-write (previously used seek→dump→truncate which
+    left a truncated file on failure).
     """
     import fcntl
 
-    with open(session_file, "r+") as f:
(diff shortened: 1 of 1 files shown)


lewtun:

Issue #345, opened 2026-06-30
Bug: session_terminated event is emitted but never surfaced in CLI or Web UI, causing silent agent death


Ojas025:
`agent/core/agent_loop.py` emits Event("session_terminated", ...) when context compaction fails, however this event type is not handled in any consumer:
- `agent/main.py` contains no case for "session_terminated" and falls through to the silent catch-all
- No existing event type definition for "session_terminated" in `frontend/src/types/events.ts`

The event is emitted correctly, the gap being purely in the consumer layers.

lewtun:
Thanks for contributing to ML Intern. The project, including its CLI, has been retired and this repository is being archived. We will therefore not be able to review or merge this work. For continued AI assistance, please use [HuggingChat](https://huggingface.co/chat/).

PR #351, opened 2026-07-15
Fix: Surface session_terminated event in CLI and Web UI

VectorCipher:
Fix: Issue #345
This PR fixes a critical bug where a context compaction failure would emit a session_terminated event that was completely unhandled by consumer layers, resulting in a silent hang/death for both CLI and Web UI users.

This PR implements the missing event handlers across the stack to ensure the agent fails loudly and gracefully.

Changes Made:
1. Python CLI (agent/main.py):
a) Added an explicit handler for session_terminated to stop the shimmer, print the user-facing error message, and correctly                       unblock the event loop.
b) Regression Guard: Replaced the silent # Silently ignore other events fallback with a logger.warning to ensure any future unhandled event types are surfaced in the logs instead of failing silently.
2. Frontend Types (frontend/src/types/events.ts):
a) Added 'session_terminated' to the EventType union for end-to-end type safety.
3. Frontend Transport (frontend/src/lib/sse-chat-transport.ts & useAgentChat.ts):
a) Added the session_terminated case to the SSE event switch statement.
b) Connected the event to the existing onError and onProcessingDone side-channels so the Web UI now correctly drops out of the loading state and displays the terminal error banner to the user.
4. Tests:
a)Added backend tests (test_session_terminated_dispatch.py) to verify the CLI explicitly handles the event and that unknown events trigger the new warning guard.
5. Testing Instructions:
1. Verified backend tests pass successfully.
3. Verified ruff linting and formatting.
4. Verified frontend TypeScript compiler exhaustiveness checks pass.

### DIFF
--- a/agent/main.py
+++ b/agent/main.py
@@ -846,7 +846,22 @@ async def event_listener(
                 )
                 await submission_queue.put(approval_submission)
                 console.print()  # spacing after approval
-            # Silently ignore other events
+            elif event.event_type == "session_terminated":
+                shimmer.stop()
+                stream_buf.discard()
+                data = event.data or {}
+                user_message = data.get(
+                    "user_message", "Session terminated unexpectedly."
+                )
+                reason = data.get("reason", "unknown")
+                print_error(f"{user_message} (reason: {reason})")
+                turn_complete_event.set()
+            else:
+                logger.warning(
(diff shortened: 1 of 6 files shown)


lewtun:

PR #359, opened 2026-08-02
Add Atlas Cloud as a direct model provider

binyangzhu000-sudo:
## Summary

- add an `atlas/<model-id>` route for the Atlas Cloud OpenAI-compatible API
- resolve `ATLASCLOUD_API_KEY` and optional `ATLASCLOUD_BASE_URL` without sending Atlas requests through HF Router
- validate and probe Atlas model switches, and avoid requiring an HF token when local tools are used

## Validation

- `uv run --extra dev ruff check` on all changed Python files
- `uv run --extra dev pytest -q tests/unit --maxfail=1` (515 passed)
- `uv run python -m compileall -q agent`
- live `deepseek-ai/deepseek-v4-pro` forced tool-call smoke test through the new resolver

### DIFF
--- a/agent/core/direct_models.py
+++ b/agent/core/direct_models.py
@@ -1,0 +1,35 @@
+"""Helpers for direct OpenAI-compatible cloud model ids."""
+
+DIRECT_MODEL_PROVIDERS: dict[str, dict[str, str]] = {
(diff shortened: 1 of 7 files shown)


lewtun:

PR #360, opened 2026-08-24
Update trl example paths in tool prompts after examples/ reorg

qgallouedec:
huggingface/trl#6820 deletes examples/scripts/, so two tool prompt strings pointed agents at a path that will 404:

- github_find_examples.py: the usage example now shows a real post-reorg hit (examples/sft_gemma3/sft_gemma3.py)
- research_tool.py: the github_read_file example now reads trl/scripts/sft.py, the CLI script that replaced examples/scripts/sft.py

Tool logic is untouched, it searches generic examples/ and scripts/ dirs and keeps working after the reorg.


lewtun:

PR #364, opened 2026-09-14
Support OpenAI-compatible gateways, and actually read the project .env

Strongich:
Four related fixes for running `ml-intern` against an OpenAI-compatible LLM gateway instead of HF Router -- to use corporate LLM gateways.

## 1. `.env` is read from the launch directory

`load_config()` does:

```python
load_dotenv(_PROJECT_ROOT / ".env")
load_dotenv(override=False)
```

The comment above it says "then CWD .env can override if present", but that isn't what happens. python-dotenv's `find_dotenv()` defaults to `usecwd=False`, which starts the search at **the calling module's directory** — `agent/`— not the process CWD. So once the CLI is installed (`uv tool install -e .`), the second call just walks up from the install root and finds the same `.env` as the first. A `.env` in the user's own project is never read.

Fixed by passing `usecwd=True` for the launch-directory lookup and keeping the install-root `.env` as a fallback, so the priority now is:
OLD: shell -> install .env
NEW: shell -> launched project .env -> install .env

Both calls stay `override=False`, so variables exported in the real environment still win; the only change in precedence is that the launch directory now beats the install root (IMO that's the most optimal approach instead of adding corporate .env params into your local fork that u can forget about later).

The regression test fails on `main` with `KeyError: 'ML_INTERN_DOTENV_PROBE'` and passes with the fix.

## 2. `openai-compat/` prefix for self-hosted and corporate gateways

`openai-compat/` was already reserved in `RESERVED_LOCAL_MODEL_PREFIXES` and rejected with "Unsupported local model id". This implements it, since a gateway is exactly the case the local prefixes can't express: today the only way to reach one is to misuse `vllm/` and point `VLLM_BASE_URL` at something that is neither vLLM nor local.

```bash
OPENAI_COMPAT_BASE_URL=https://llm-gateway.your-company.com/openai/v1
OPENAI_COMPAT_API_KEY=<your-gateway-key>

ml-intern --model openai-compat/gpt-5.5 "your prompt"
ml-intern --model openai-compat/vertex/claude-opus-4-7 "your prompt"
```

It reuses the existing local-provider plumbing (same base-URL/API-key env pair, same `/v1` normalisation, same `openai/<model>` handoff to LiteLLM), with two differences:

- **No default base URL.** There's no sensible localhost guess for a gateway, so an unconfigured one raises `No base URL configured for 'openai-compat/…'. Set OPENAI_COMPAT_BASE_URL (or LOCAL_LLM_BASE_URL) …` rather than quietly dialling `localhost`.
- **Multi-segment model names pass through.** Gateways namespace models by vendor (`vertex/claude-opus-5` in my case), so only the prefix is stripped.

These ids no longer need `HF_TOKEN` just to start a session -- inference is billed to the gateway. `HF_TOKEN` is still required for the Hub-backed tools and `--sandbox-tools`.

`RESERVED_LOCAL_MODEL_PREFIXES` / `is_reserved_local_model_id` are removed, as `openai-compat/` was their only member. If you suggest keeping it -- say so, happy to change.

## 3. Gateway ids are remote, not local, so `/effort` keeps working

My first pass just added `openai-compat/` to `LOCAL_MODEL_PROVIDERS`, and that turned out wrong. `is_local_model_id` was doing two unrelated jobs -- "skip the HF token prompt" and "skip the effort probe" -- so every gateway model printed `checking local model …` and committed with `effort: off`. `/effort high` was accepted and then silently dropped. A gateway fronts the same frontier models HF Router does, so that's not the right default.

Split into three predicates:

```python
is_local_model_id            # the 4 localhost servers -- no thinking params
is_openai_compat_model_id    # gateways -- full effort probe
is_direct_endpoint_model_id  # either -- skips HF token + router catalog
```

Gateways now forward `reasoning_effort` and go through the same `probe_effort` cascade as router ids, so whatever the far side accepts gets discovered instead of assumed -- no capability table needed. Local prefixes still drop effort, and there's a test pinning that the carve-out doesn't leak into them.

`backend/routes/agent.py`'s `_model_requires_hf_router_token` moves to the new umbrella predicate too.

## 4. Show the resolved endpoint

With three possible sources for `OPENAI_COMPAT_BASE_URL` (shell, launch-dir `.env`, install-root `.env`) and per-variable precedence, a base URL and an API key can come from different files -- a combination nobody actually configured. My gateway reports that as `not allowed for this virtual key`, which sends u looking at key permissions instead of at config. So the banner, headless startup and `/status` now print where a direct-endpoint id resolves to:

```
Model: openai-compat/claude-opus-4-7
Endpoint: https://bifrost.infra.reface.rocks/openai/v1
Tool runtime: local filesystem
```

Omitted for HF Router ids, which always go to the same place. The line sits above `Tools: loading...` on purpose -- `print_init_done` walks back a fixed number of rows to rewrite that line, so anything below it would break the rewrite.

## Testing

- `uv run ruff check .`, `uv run ruff format --check .` and `uv run pytest tests/unit` (528 passed).
- Verified end-to-end against a real corporate Bifrost gateway: credentials read from the project `.env` with nothing exported in the environment, `--model openai-compat/vertex/claude-opus-5`, full agent turn completed. Effort probe resolves live against the gateway:

```
checking openai-compat/vertex/claude-opus-5 (effort: high)...
Model switched to openai-compat/vertex/claude-opus-5 (effort: high, 3001ms)
```

image

As you can see, there's some upstream "overlay" of the MCP connection message and `/quit to exit`. That's an existing `main` issue, not from this PR, so I addressed them both in 2 separate PRs:
1. https://github.com/huggingface/ml-intern/pull/365 -- boot banner fix
2. https://github.com/huggingface/ml-intern/pull/366 -- fix MCP connector


Strongich:
Sorry, the agent fixed my inquiry and decided to open a PR by himself

